PR #61, opened 2014-10-15
Added rapture-core project to community build

propensive:
Is this all that's required to include a new project in the build? Can someone let me know if this works, and if so I'll add the other Rapture projects? Thanks!


SethTisue:
successor: #203

PR #65, opened 2014-10-20
Add basic build for Scala.js.

sjrd:
I did this in "blind" mode. I cannot test it locally due to #64.

PR #94, opened 2015-03-04
Resolve ambiguity in scala.js's command.

gkossakowski:
It's an import ambiguity that we need to fix.


gkossakowski:
I broke community build by merging #65. This should fix the problem. It has been tested here already:
https://jenkins-dbuild.typesafe.com:8499/job/Community-2.11.x-manual/143/

PR #107, opened 2015-06-29
expand readme considerably

SethTisue:
review/merge by @adriaanm


adriaanm:
LGTM! The config is not super friendly to running fully locally, as it uses the scala-ci caching proxy (which caches a ton of different repos -- see comments in the conf file), which speeds up builds considerably when running close to that server, but not necessarily when running somewhere else.

PR #123, opened 2015-07-30
unfork scalaz-stream

SethTisue:


SethTisue:
verified good by https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/105/console

PR #119, opened 2015-07-22
2.11.x failing due to missing spray-json-shapeless dependency for ensime-server

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/49/console shows the failure:

> Missing dependency: the library com.github.fommil#spray-json-shapeless is not provided (in space "default") by any project in this configuration file`

the cause is https://github.com/ensime/ensime-server/commit/88cd6e33e946c472e1a1c3477942425211ac6974 

perhaps it will be straightforward to just add the dependency? we'll see. spray-json-shapeless's dependencies are here: https://github.com/fommil/spray-json-shapeless/blob/master/build.sbt

PR #126, opened 2015-08-25
freeze ensime rev for now

SethTisue:
we freeze at a rev from before spray-json-shapeless was added as a
dependency, since that would set off a cascade of upgrades


SethTisue:
green build with this change at https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/65/console

SethTisue:
to see what I mean by “cascade of upgrades”, see the (unmerged) commits in #119

PR #172, opened 2015-11-10
remove unused file

SethTisue:
review by @adriaanm. if this might havesome value in the future, I
suggest we make an issue documenting that (and linking to the commit
with the removal), rather than just leaving the file lying around.


adriaanm:
LGTM -- cleanliness ftw

PR #184, opened 2015-11-19
use triple quotes to make appendScalacOptions more readable

SethTisue:
it looks like whoever put this in didn't know HOCON supported
triple quotes. neither did I until today.


SethTisue:
oops'ed in ad9a0aa36ce7465e2a64e5b9ac0401861ac2d511

PR #186, opened 2015-11-20
fix appendScalacOptions, unfork scala-js

SethTisue:
green run at
https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/161/consoleFull

well, it's still running, but scala-js is green, and I see the right
commands in the dump at the top of the run. works for me


SethTisue:
(approach here w/r/t avoiding `+=` is the same as used in RP build)

PR #189, opened 2015-11-24
upgrade to dbuild 0.9.4

SethTisue:
we had held back over concerns with +=, but since 50544fd that is
no longer an issue


SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.11.x-jdk8-integrate-community-build/106/consoleFull

SethTisue:
@cunei @eed3si9n it might interest you both to know that we're using dbuild 0.9.4 now — to be on the same page with y'all and with RP

PR #194, opened 2015-12-04
upgrade dbuild to 0.9.5

SethTisue:
mainly just to keep current w/ what RP is doing, but also 0.9.5
has multi-JDK support so we should be able to remove some JDK
6 vs. 8 workarounds as this is merged forwards


SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/168/consoleFull

PR #199, opened 2015-12-15
allow passing ant options when building scala/scala

lrytz:


lrytz:
companion of https://github.com/scala/scala-jenkins-infra/pull/147, build running here: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/269/parameters/. the Scala build picked up the option.

lrytz:
I just realize I'd also need an option to do a full bootstrap (disalbe `locker.skip` in https://github.com/scala/community-builds/blob/2.11.x/community.dbuild#L35)

SethTisue:
nbd, but may I suggest calling it `scala-build-opts` instead of `scala-ant-opts`, in anticipation of the ant->sbt switchover?

lrytz:
sure, that's fine with me. dbuild will require some other adjustments when we do the switch, and the option introduced here will not continue to work in the same way.

lrytz:
@SethTisue PTAL, I didn't find a better way than the current form.

Issue #202, opened 2016-01-14
Shapeless build fails in recent 2.11.x builds


SethTisue:
```
[shapeless] [error] **** Missing dependency: the library org.scalamacros#paradise is not provided (in space "default") by any project in this configuration file.
[shapeless] [error] In order to control which version is used, please add the corresponding project to the build file
[shapeless] [error] (or use "check-missing:false" to ignore (not recommended)).
[shapeless] java.lang.RuntimeException: Required dependency not found
```
- last 2.11.x-jdk6 success: d6a4f56e0783655c48244ec23c4f4c2d98401a72 (Jan 4)
- first 2.11.x-jdk6 failure: 7c0a87364c5d990fc50923dfd0af8aaada4683c2 (Jan 14)
- last 2.11.x-jdk8 success: same as above  
- first 2.11.x-jdk8 failure: 2ce5d7dfa1fac107b277bee862708ff0c9ac71ca (Jan 13)

```
% git hist 2ce5d7dfa1fac107b277bee862708ff0c9ac71ca ^d6a4f56e0783655c48244ec23c4f4c2d98401a72
* 2ce5d7d - Revert "Updated Sonatype realm." (26 hours ago) <Miles Sabin>
*   8075ac4 - Merge pull request #531 from dwijnand/non-fatal-console (26 hours ago) <Miles Sabin>
|\  
| * 454c51a - Disable -Xfatal-warnings in console (28 hours ago) <Dale Wijnand>
|/  
*   23ace5c - Merge pull request #528 from dwijnand/CompatContext-is-a-runtime.Context (34 hours ago) <Miles Sabin>
|\  
| * 07f4aba - Drop c2 in CaseClassMacrosMixin (compat.scala) (2 days ago) <Dale Wijnand>
|/  
* 495e275 - Updated Sonatype realm. (2 days ago) <Miles Sabin>
* 0e2a141 - Refreshed Travis env vars. (2 days ago) <Miles Sabin>
* d16892e - Merge pull request #515 from dwijnand/uni (2 days ago) <Miles Sabin>
* e387788 - Add myself (Dale) to compat.scala copyright holders (2 days ago) <Dale Wijnand>
* 93413fc - Switch to contiguous copyright date ranges (2 days ago) <Dale Wijnand>
* aede308 - Add macro-compat to enable Scala 2.10 in master (3 days ago) <Dale Wijnand>
* 81274d7 - Misc cleanups (3 days ago) <Dale Wijnand>
```

[...]

PR #204, opened 2016-01-14
freeze Shapeless for now

SethTisue:
#202 explains why


SethTisue:
reverted in #206

milessabin:
Thanks :-)

PR #208, opened 2016-01-21
add scodec-bits

SethTisue:
we have it in the 2.12 build, can't see any reason it shouldn't
be in 2.11 as well

### DIFF
--- a/common.conf
+++ b/common.conf
@@ -107,6 +107,7 @@ vars: {
   sbinary-ref                  : "harrah/sbinary.git"
   sbt-republish-ref            : "typesafehub/sbt-republish.git"
   scalaz-ref                   : "scalaz/scalaz.git#series/7.1.x"
+  scodec-bits-ref              : "scodec/scodec-bits.git#series/1.0.x"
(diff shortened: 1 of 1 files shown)


SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/208/ (link will 404 until job reaches head of queue)

SethTisue:
```
[scodec-bits] --== Extracting dependencies for scodec-bits ==--
...
[scodec-bits] java.lang.UnsupportedClassVersionError: pl/project13/scala/sbt/JmhPlugin$ : Unsupported major.minor version 51.0
[scodec-bits]   at java.lang.ClassLoader.defineClass1(Native Method)
```

51.0 is Java 7, we're on Java 6. but JMH is only a dependency of the `benchmark` subproject, which we don't need to include.

SethTisue:
ah, but it's a plugin, I missed that JMH is actually a transitive dependency of the scodec-build plugin. so I'll close this PR and make a new one for the 2.11.x-jdk8 build instead.

PR #209, opened 2016-01-21
add scodec-bits

SethTisue:
we have it in the 2.12 build, can't see any reason it shouldn't
be in 2.11 as well. adding it to 2.11.x-jdk8 rather than regular
2.11.x because it uses an sbt plugin (namely scodec-build) with
a Java 7+ only dependency (namely JMH)

### DIFF
--- a/common.conf
+++ b/common.conf
@@ -111,6 +111,7 @@ vars: {
   sbinary-ref                  : "harrah/sbinary.git"
   sbt-republish-ref            : "typesafehub/sbt-republish.git"
   scalaz-ref                   : "scalaz/scalaz.git#series/7.1.x"
+  scodec-bits-ref              : "scodec/scodec-bits.git#series/1.0.x"
   discipline-ref               : "typelevel/discipline.git#v0.2"
   scala-js-ref                 : "scala-js/scala-js.git"
   slick-ref                    : "slick/slick.git"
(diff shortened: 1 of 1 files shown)


SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.11.x-jdk8-integrate-community-build/132/consoleFull (404 til the worker boots up)

SethTisue:
failed — looks like we probably just need to disable `-Xfatal-warnings`. I see in 2.12.x common.conf we have custom scalacOptions; I guess that's why. (I was a bit mystified by this til I realized the `-Xfatal-warnings` is actually coming from the scodec-build sbt plugin, not from scodec-bits' own .sbt/.scala files)

We should be able to remove the unwanted option more surgically now that we have `removeScalacOptions` in our toolbox.

SethTisue:
new test run: https://scala-ci.typesafe.com/job/scala-2.11.x-jdk8-integrate-community-build/133/consoleFull

SethTisue:
tests failed to compile, so I added:

``` diff
+    // core/shared/src/test/scala/scodec/bits/BitsSuite.scala:9:
+    //   not found: type PropertyCheckConfiguration
+    // looks like probably a ScalaTest 2.x vs 3.0 thing to me
+    run-tests: false
```

new test run: https://scala-ci.typesafe.com/job/scala-2.11.x-jdk8-integrate-community-build/134/consoleFull (404 til worker spins up)

PR #231, opened 2016-03-03
2.12: jawn, spire: don't use removed -Yinline-warnings flag

SethTisue:
jawn and spire started failing in the 2.12.x build
when `-Yinline-warnings` was removed from 2.12

### DIFF
--- a/common.conf
+++ b/common.conf
@@ -751,6 +751,10 @@ build += {
     // omitted TODO: play
     // omitted: argonaut, rojoma-v3, rojoma, benchmark
     extra.projects: ["ast", "parser"]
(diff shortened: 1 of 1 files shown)


SethTisue:
test run: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/332/consoleFull

@non in both projects, I think you'll want to make the use of this flag conditional on `scalaVersion`. (if you do that, I'll be able to back out the Jawn part of this PR, but the Spire part will have to stay for now, since we are building Spire off an old SHA)

PR #237, opened 2016-04-04
freeze utest for now

SethTisue:


SethTisue:
bah. this version is very old (March 2015). better to temporarily patch acyclic

PR #241, opened 2016-04-06
fix sbinary

### LINKED ISSUE #240: sbinary won't build in 2.12 because `Predef.error` was removed
also we're currently building harrah/sbinary, but Mark isn't active on GitHub anymore afaict, so I should investigate whether we should switch over to sbt/sbinary, which appears to be more maintained

SethTisue:
fixes #240


SethTisue:
test run: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/369/console

PR #247, opened 2016-07-12
Add https://github.com/scala-js/scala-2.12-junit-mixin-plugin

sjrd:
This is used by Scala.js and Shapeless (at least) with Scala 2.12.


adriaanm:
Super, thank you!

PR #267, opened 2016-08-31
add Scalactic

SethTisue:
this is good to have in general, but also Akka depends on it now (in a
part of Akka we aren't currently building, hopefully will be soon)


SethTisue:
already green in several recent test runs

SethTisue:
fyi @bvenners

PR #275, opened 2016-09-05
Resolve scala from maven

adriaanm:
So we can [drop ant](https://github.com/scala/scala/pull/5311)


adriaanm:
closing in favor of Seth's polished version

PR #286, opened 2016-09-19
Use "nodejs" in all Scala.js projects that testSuite depends on.

sjrd:
Although only `testSuite` actually has tests, dbuild will try to run the tests for all projects that `testSuite` depends on (transitively), so we need to set it in a bunch of places.


sjrd:
Review by @SethTisue

SethTisue:
let's try it https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/308/consoleFull

SethTisue:
green

PR #295, opened 2016-09-23
play-doc

schmitch:
I guess that should do it?
How does it pick up 2.12 Dependencies?


SethTisue:
> I guess that should do it?

Let's try it: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/623/consoleFull

> How does it pick up 2.12 Dependencies?

dbuild will look into play-doc's build, see what Scala-based dependencies it has, and rewire them to point to the versions that were built during the current run

SethTisue:
the fastparse failure is unrelated, so, merging. thank you!

PR #298, opened 2016-09-23
add akka-http & akka-stream

### LINKED ISSUE #146: add Akka HTTP and Akka Streams
we should do this because it's Akka, but also because ensime-server's master branch now has them as a dependency

I tried changing our dep to `akka.git#releasing-akka-stream-and-http-experimental-1.0` but it didn't work and I decided to postpone rather than investigate. (for now we freeze ensime-server at the commit before the dependency was added)

(this was in the 2.11 build. it might be better to leave that one alone and only tackle Akka Streams in Java 8 world? dunno)

### LINKED ISSUE #292: add ssl-config to community build (needed by akka-http)
needed by akka-http, which I'm trying to build by adding `"akka-http-core", "akka-http-tests"` to extra.projects in the "akka" entry in common.conf

SethTisue:
also adds akka-cluster, akka-remote-tests, ssl-config

fixes #146, fixes #292

green run at
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/625/console

### DIFF
--- a/common.conf
+++ b/common.conf
@@ -582,6 +582,7 @@ build += {
         options: ["-Dakka.genjavadoc.enabled=false", "-Dakka.scaladoc.diagrams=false", "-Dakka.build.aggregateSamples=false"]
         projects: ["akka-scala-nightly"]
         exclude: [
+          "akka-docs"   // this is Sphinx stuff, not really apropos here, no Sphinx on Jenkins anyway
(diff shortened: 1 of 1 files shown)


SethTisue:
sorry for the slightly weird git history, not worth cleaning up IMO

PR #312, opened 2016-09-27
add cats-kernel and simulacrum

SethTisue:
we can add more of cats later. (when we do, we'll need simulacrum.)

fyi @non @mpilquist @BennyHill


SethTisue:
passing locally

PR #320, opened 2016-09-29
cleaner structure for config file

SethTisue:
now that the 2.11 builds have moved over to light-maintenance mode,
we're free to let this one diverge.  merging changes from the old
builds will have to be done manually, but that's fine.

I believe, based on my long experience with the old format, that this
new format will be easier to maintain.

(eventually I plan to go on and put each project in its own file)

review by @adriaanm


SethTisue:
green run: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/693/consoleFull

adriaanm:
I like the simplifications. I don't think common.conf is a good name, though :-) git doesn't care about file names, just content, so it shouldn't matter in which direction you consolidate.

It was kind of nice to have a top-level overview of which community projects we had to fork, but maybe we can actually do even better in the new format by having a naming convention for the branches that constitute these forks, or by creating a github user that has the forked repos for the community build, so we have a centralized spot to track our forks.

SethTisue:
> I like the simplifications. I don't think common.conf is a good name, though :-) git doesn't care about file names, just content, so it shouldn't matter in which direction you consolidate.

fixed

> It was kind of nice to have a top-level overview of which community projects we had to fork

I'm fine with `grep uri: community.dbuild`

(the effort to keep the forked and unforked projects in separate sections was never better than half-hearted, and it made diffs/merging/history super annoying)

> but maybe we can actually do even better in the new format [...]

https://github.com/scala/community-builds/issues/321

> creating a github user that has the forked repos for the community build

https://github.com/scala/community-builds/issues/322

PR #339, opened 2016-10-13
give Spire more heap

SethTisue:
see commit message for justification


SethTisue:
@adriaanm (this [hopefully] addresses the spire failure in your current run)

Issue #338, opened 2016-10-12
freeze all SHAs, move them forward together weekly?


SethTisue:
we track master of many projects now, especially in the 2.12 build

that's **good** because
- we find out about problems quickly

but it's also **bad** because
- anytime something fails, we never know right away whether it's because Scala changed, or because the project changed
- it slows down maintenance of the community build, because you can be in the middle of testing some change, when suddenly some cornerstone project like ScalaTest or Specs2 merges a trivial PR, and now everything has to be rebuilt and the change you thought would take 10 minutes to test suddenly takes 8 hours

**proposed fix**:
- freeze all projects to fixed SHAs
- write a script that bumps all the SHAs to the latest
  - with some mechanism for leaving certain projects frozen, perhaps a special comment on that line
- ...and submits a pull request and starts a test run
- run the script weekly or so
- if the test run is green, merge the PR, if not, investigate

**or maybe**: it would also be possible to have separate "stable" and "bleeding edge" community builds, but I'm not sure maintaining them and keeping them in sync and making the necessary Jenkins changes in Chef is worth the effort. I think the above scheme might strike a nice cost/benefit balance

dwijnand:
I think your proposal is very viable.

But I wonder if perhaps the opposite makes sense - given:
- the projects in community build change much more frequently than scala/scala
- any reasonably aggressive change in scala/scala is tested with the community build even before merging

Why not freeze scala/scala instead, and bump at some interval (nightly/weekly/etc)?

adriaanm:
Here's how I'm implementing this:

To turn the repeatable build json into a list of URIs, extract the JSON between

```
[info] You can repeat this build (except for -SNAPSHOT references) using this configuration:
```

and

```
[info] ---== End Repeatable dbuild Configuration ===---
```

```
curl -s \
https://scala-ci.typesafe.com/view/scala-2.12.0/job/scala-2.12.0-integrate-community-build/17/consoleText \
| awk '/.*You can repeat this build.*using this configuration:/{y=1;next}/.*---== End Repeatable dbuild Configuration ===---/'{y=0}y \
| jq -re '.build.projects[] | "\(.name)-uri: \"\(.uri)\""' | grep -v scala-uri | sort
```

First shot: https://github.com/scala/community-builds/compare/2.12.x...adriaanm:2.12.x?expand=1
Validation:  https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/796/console

milessabin:
[...]

PR #343, opened 2016-10-13
Jenkins friendliness, set stage for frozen SHAs

adriaanm:
For future reference,  to turn the repeatable build json into a list of URIs, extract the JSON between

```
[info] You can repeat this build (except for -SNAPSHOT references) using this configuration:
```

and

```
[info] ---== End Repeatable dbuild Configuration ===---
```

save it as rep.json and:

```
cat rep.json | jq -re '.build.projects[] | "\(.name)-uri: \"\(.uri)\""'
```

See #338


adriaanm:
I'm going to merge this to allow progress with the 2.12.0 branch  / avoid breaking jenkins (param name change)

PR #360, opened 2016-10-20
fix regression: MiMa failing to extract

SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.11.x-jdk8-integrate-community-build/245/consoleFull (404 til Jenkins gets in gear)

### DIFF
--- a/common.conf
+++ b/common.conf
@@ -575,6 +575,8 @@ build += {
     // macros compiled by previous versions of Scala" errors I don't want
     // to take the time to investigate
     extra.exclude: ["sbtplugin"]
+    // the build was recently redone and no longer works with 0.13.12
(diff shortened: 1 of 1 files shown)


SethTisue:
@dwijnand you might want to look into the details here, this is just me trying for the quick fix without really investigating

dwijnand:
extraProjects is a 0.13.13 feature, so that's the reason.

Did it fail during extracting or running? Because your test run above extracted fine.

SethTisue:
It failed during extraction before. aborting the run and merging.

dwijnand:
In retrospect we shouldn't've upgraded mima to the prerelease

PR #367, opened 2016-10-25
update cats entry

SethTisue:
now that they merged 2.12 support

note that we are still using a fork, because of the POM issue (see
comment). but I did also bring the fork up to date with their current
master

fyi @non @adelbertc

### DIFF
--- a/community.dbuild
+++ b/community.dbuild
@@ -411,17 +411,15 @@ build += {
     ]
   }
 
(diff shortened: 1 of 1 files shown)


SethTisue:
passing locally. merging

adelbertc:
POM issue PR: https://github.com/typelevel/cats/pull/1426

PR #378, opened 2016-10-28
upgrade to sbt 0.13.13 final

SethTisue:


SethTisue:
optimistically merging, since rebuilding everything takes so long. if there turns out to have been some regression between 0.13.13-RC3 and 0.13.13, we'll deal with it

PR #381, opened 2016-10-28
[2.12.0] Upgrade to sbt 0.13.13

dwijnand:


SethTisue:
I would suggest we leave the 2.12.0 build where it is. Everything is frozen there.

PR #353, opened 2016-10-18
compile many more tests

SethTisue:
also: re-add scalariform
also: drop scala-refactoring until
  https://github.com/scala-ide/scala-refactoring/issues/168
  is addressed

fixes #280

Issue #280, opened 2016-09-11
make sure we are compiling tests


SethTisue:
even when we don't run them

in the run-up to 2.12 it happened at least twice that we missed out finding out about regressions because of this:
- scala-refactoring: https://github.com/scala/scala/pull/5388#issuecomment-246170392 (@Sciss)
- shapeless: https://github.com/scala/scala/pull/5394 (@milessabin)

SethTisue:
dbuild ticket: https://github.com/typesafehub/dbuild/issues/189 /cc @dwijnand

SethTisue:
I did some local experimentation, and it turns out to work fine to replace `run-tests: false` with `extra.test-tasks: ["compile"]` if we want to compile tests without running them. I will make a PR that does that for as many projects as possible. And I've added a note about it to https://github.com/scala/community-builds/wiki/Maintenance

SethTisue:
PR #353

PR #387, opened 2016-11-08
Run Scoverage tests.

gslowikowski:
This change should not be enough. I want to see the result of first, failing, build before fixing.

### DIFF
--- a/community.dbuild
+++ b/community.dbuild
@@ -200,7 +200,6 @@ build += {
     uri:  ${vars.uris.scoverage-uri}
     extra: ${vars.base.extra} {
       exclude: ["scalac-scoverage-runtimeJS"] // no Scala.js please
-      test-tasks: "compile" // TODO: [info] java.io.FileNotFoundException: Could not locate [~/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.11.0.jar].
     }
   }
 


gslowikowski:
Can I trigger the build?

SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/921/consoleFull (404 til Jenkins finishes waxing his mustache)

SethTisue:
@gslowikowski the scoverage part of the log begins with `[scoverage] --== Building scoverage ==--`

gslowikowski:
There are errors:

```
[scoverage:error] java.io.FileNotFoundException: Could not locate [/home/jenkins/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.12.1-10c609e-nightly.jar].
```

Is `/home/jenkins/.ivy2/local/org.scala-lang/scala-compiler/jars/scala-compiler-2.12.1-10c609e-nightly.jar` the right location (`local` instead of `cache`)?

SethTisue:
@cunei is this something you could help @gslowikowski troubleshoot...?

cunei:
I am not sure what is the aim of this change, @gslowikowski can you please clarify? The "test-tasks" option is a list of tasks that are run after compilation completes successfully. If compilation does not succeed, the test tasks are not invoked at all, and dbuild will say so. By default, test-tasks is just "test"; the option exists in case there are further tests to be run in different configurations, for instance "it:test" or other kinds of tests. "compile" should not be part of "test-tasks", and most likely fails since the environment is not properly set up for a second compilation. But, as I mentioned, the test-tasks are only called after compilation completes successfully.

SethTisue:
@cunei we use `test-tasks: ["compile"]` in cases where we can successfully compile the tests, but expect them to fail if they were actually run, so we don't run them. see #280 and #353 (and also https://github.com/typesafehub/dbuild/issues/189)

this PR removes that line for this one project; it's an attempt to actually start running the tests and see what happens. and what happens is the `FileNotFoundException`s referenced above. and that's what we're hoping to troubleshoot.

gslowikowski:
I will describe you tomorrow, how Scoverage test harness works.

cunei:
@SethTisue I see, thanks. I never thought that would be a possible usage, I'm actually a bit surprised it works at all. Shouldn't it be "test:compile"? I don't think I explicitly scope the tasks to the "test" configuration in dbuild, I need to check.
@gslowikowski, thanks! The list of resolvers and the dependencies are patched by dbuild before running the various tasks; even the ivy home is moved elsewhere. If the harness spawns a new process, it may be that the default settings are used, causing the newly compiled, and rewritten, project dependencies not to be found.

cunei:
@SethTisue, I check the scoping rules: if the task name contains a configuration, like "config:task", then that configuration is used; if just the task name is specified, then the "test" configuration is used by dbuild. Which is why it works :) I'll add that to the docs.

gslowikowski:
Scoverage test harness needs to [prepare classpath](https://github.com/scoverage/scalac-scoverage-plugin/blob/v1.3.0/scalac-scoverage-plugin/src/test/scala/scoverage/ScoverageCompiler.scala#L20) for compiler tests.

Classpath contains: `scala-compiler`, `scala-library`, `scala-reflect` ([see](https://github.com/scoverage/scalac-scoverage-plugin/blob/v1.3.0/scalac-scoverage-plugin/src/test/scala/scoverage/ScoverageCompiler.scala#L45-L48)) and `scala-logging` dependencies for some macro tests ([see](https://github.com/scoverage/scalac-scoverage-plugin/blob/v1.3.0/scalac-scoverage-plugin/src/test/scala/scoverage/ScalaLoggingSupport.scala#L20-L34)).

I the future we will try to get rid of `scala-logging` dependency somehow.

Now all dependencies are expected to be in present in `.ivy2/cache` directory ([see](https://github.com/scoverage/scalac-scoverage-plugin/blob/v1.3.0/scalac-scoverage-plugin/src/test/scala/scoverage/ScoverageCompiler.scala#L61-L69)). This code does not work in `dbuild`.

One improvement should be checking `ivy2/local` before `.ivy2/cache` for e.g. locally installed Scala snapshots.

Second improvement - compatibility with [SBT scalaHome feature](http://www.scala-sbt.org/0.13/docs/Configuring-Scala.html#Using+Scala+from+a+local+directory). The main question is, how can I check Scala home. Should I use `scala.util.Properties.scalaHome` (just found it)?

gslowikowski:
I'm trying to run minimal dbuild locally (Windows XP) basing on community build script, but `dbuild` calls module builds with empty command, like:

```
cmd: List(
 java,
 -Dsbt.global.base=J:\scov.dbuild2\.\target-0.9.7-RC1\sbt-base-dir\extractor,
 -Dsbt.version=0.13.8,
 -Dsbt.override.build.repos=true,
 -Dsbt.log.noformat=true,
 -Ddbuild.sbt-runner.last-msg=D:\home\gs\Temp\sbtrunner4203287758928261410lastExceptionMessage,
 -Dsbt.repository.config=J:\scov.dbuild2\target-0.9.7-RC1\extraction\94333df9af3e226a70cc6f2eb262f9ee4cf1eceb\projects\bfcc8989708b442aef58bb83eec35fdf14ca8eec\projects\6a4e84fc514f102c93c09f0a80583194dd6e8b4d\.dbuild\repositories,
 -XX:+CMSClassUnloadingEnabled,
 -Xms512m,
 -Xmx1024m,
 -Xss2m,
 -XX:ReservedCodeCacheSize=192m,
 -jar,
 J:\scov.dbuild2\.\target-0.9.7-RC1\sbt-base-dir\extractor\launcher\sbt-launch.jar,

)
```

There is empty string at the end of the list.
The build enters shell and hangs (waits for input). I tried `dbuild` versions `0.9.5`, `0.9.6` and latest `0.9.7.RC1`

On the other hand, `dbuild` configures Scala home for the build so I tried to build Scala 2.12.x locally and add `scalaHome` to Scoverage build. I cannot get Scala home in runtime to find scala files - `scala.util.Properties.scalaHome` returns empty string. What am I doing wrong?

SethTisue:
`scala.util.Properties.scalaHome` returns the empty string when I run it from `sbt console`, no dbuild involved, so apparently you will need to use some other method to find the JARs. Perhaps sbt has something? I poked around a bit, to no avail.

dwijnand:
@gslowikowski 

> dbuild configures Scala home for the build

Can you show where you found that to be the case?

dwijnand:
Looks like you're looking for the Scala jars when defining the classpath scalac's Settings. I'm no scalac plugin expert, but surely there are other compiler plugins out there that have solved this issue, without using a fragile technique like looking in `${user.home}/.ivy2/cache`.

cunei:
@dwijnand ScalaInstance is redefined by dbuild; that is necessary since we want all the projects to be compiled with the newly compiled scala compiler, if one is available.
https://github.com/typesafehub/dbuild/blob/89a970d3a2947d1ceb587bf8e0430c8b648b2440/plugin/src/main/scala/com/typesafe/dbuild/plugin/DBuildRunner.scala#L426

cunei:
@gslowikowski dbuild itself needs to find the sbt.jar file, and traverses the classpath in order to find the one in use. You may be able to use a similar technique to retrieve the scala jars, knowing the scalac version number. See:
https://github.com/typesafehub/dbuild/blob/31806a431209b016e0813965056918d69de3b1d8/support/src/main/scala/com/typesafe/dbuild/support/sbt/SbtRunner.scala#L136-L142
https://github.com/typesafehub/dbuild/blob/31806a431209b016e0813965056918d69de3b1d8/support/src/main/scala/com/typesafe/dbuild/support/sbt/SbtRunner.scala#L161-L166

SethTisue:
closing for inactivity. could be revived

PR #392, opened 2016-11-10
fix regression where zinc failed to extract

SethTisue:
they're officially on sbt 0.13.11 now, so I expect 0.13.13
will work too


SethTisue:
green run at https://scala-ci.typesafe.com/view/scala-2.11.x/job/scala-2.11.x-integrate-community-build/390/consoleFull

PR #394, opened 2016-11-12
unfork lift-json

SethTisue:
needed change was merged upstream


SethTisue:
already tested locally (against 2.12.0 anyway, since macro-paradise is broken for now on 2.12.x)

PR #395, opened 2016-11-13
[nomerge] freeze scala-gopher at last commit that works on JDK 6

SethTisue:
this is for the 2.11.x branch only; we won't keep the change on
the 2.11.x-jdk8 or 2.12.x branches

fyi @rssh


SethTisue:
doh, fixup in 80da1f8df404d4731a985ce87200400c5666c06c

PR #403, opened 2016-11-21
make local runs easier

SethTisue:
by asking Jenkins for a Scala version number if none is specified


SethTisue:
come to think of it, this works in the Jenkins UI too. (though it won't make "rebuild" build the same version again.)

SethTisue:
doh, silly me, fixup in 5b8f61b6b528f7e34228b27a59269a4813a1c336

PR #404, opened 2016-11-22
fix regression in macro paradise

SethTisue:
(the spray-json-shapeless and pcplod freezes are unrelated, but are
presently needed for extraction to succeed)


SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/961/consoleFull

I'll make a separate PR that unfreezes spray-json-shapeless and pcplod; @fommil says that the offending plugin is already on its way to Maven Central

Issue #416, opened 2016-12-01
2.12.x regression: play2-core failing


SethTisue:
a big 2.12 support commit (of @wsargent's) dropped recently, so maybe it has something to do with that.

there's an off chance there's a Scala regression here that's a concern for the 2.12.1 release

SethTisue:
at e.g. https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1010/consoleFull we see:

```
[play2-core] [info] Compiling 5 Scala sources and 7 Java sources to /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.7-RC1/project-builds/play2-core-633ae55f7795a692763d273744261730c41cb13f/framework/src/play-test/target/scala-2.12.1-50d46c1-nightly/classes...
[play2-core] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.7-RC1/project-builds/play2-core-633ae55f7795a692763d273744261730c41cb13f/framework/src/play-test/src/main/scala/play/api/test/Helpers.scala:468: object creation impossible, since method materialize in class Materializer of type [Mat](runnable: akka.stream.Graph[akka.stream.ClosedShape,Mat], initialAttributes: akka.stream.Attributes)Mat is not defined
[play2-core] [error] private[play] object NoMaterializer extends Materializer {
[play2-core] [error]                      ^
[play2-core] [error] one error found
[play2-core] [error] (Play-Test/compile:compileIncremental) Compilation failed
[play2-core:error] 
```

maybe an Akka version mismatch?

SethTisue:
maybe freezing Akka at the `v2.4.12` tag will fix it; that's the version of the binary dependency Play currently has. test run: ~~https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1015/consoleFull~~ https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1016/consoleFull

SethTisue:
in #417 I froze Akka at 2.4.12 to make the problem go away. but that's not a good long-term solution. @wsargent @patriknw @ktoso? did Akka inadvertently make a source-incompatible change here?

ktoso:
Oh, extending materializer is not covered by compat guarantees – Play should fix their `NoMaterializer` 

I can send in a PR for that.

ktoso:
And done: https://github.com/playframework/playframework/pull/6767

[...]

PR #417, opened 2016-12-01
freeze Akka at 2.4.12 for now

SethTisue:
so Play will compile. works around
https://github.com/scala/community-builds/issues/416


SethTisue:
green test run at https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1016/consoleFull

PR #436, opened 2016-12-11
fix some projects that fail to extract for 2.13

SethTisue:
in the case of monix and scala-continuations, fork to get a trivial
build change in.  I opened upstream PRs for both.

in the case of parboiled 1, just drop it. we have parboiled 2.


SethTisue:
test run: https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/79/consoleFull

SethTisue:
extractomundo! everything extracted.

that's all we're aiming for at the moment.

PR #448, opened 2017-01-04
ScalaCheck 1.13!

SethTisue:
this is already green locally when run against 2.12.1, except for a
few known failures that have already been seen elsewhere, and
a spray-json-shapeless failure that is probably unrelated and
that we can deal with separately later

cc @rickynils

this PR also:

* unfreezes argonaut (cc @seanparsons @xuwei-k)
* runs lift-json's test suite again (cc @Shadowfiend)
* adds cats-laws (cc @adelbertc @non)
* adds another monix subproject (cc @alexandru)


rickynils:
Awesome!

PR #449, opened 2017-01-05
add http4s

SethTisue:
* use topic/cats branch, since master has a scalaz-stream dependency
* also add dependencies: blaze, log4s, http4s-websocket, fs2-cats, jawn-fs2


SethTisue:
I've seen this pass locally when run against 2.12.1, so, merging.

SethTisue:
* http4s: fyi @rossabaker @ChristopherDavenport
* blaze: fyi @bryce-anderson
* log4s: fyi @sarahgerweck
* fs2-cats: fyi @mpilquist

SethTisue:
@rossabaker it's bit sad I had to disable running tests because of the `org.http4s.server.middleware.DefaultHeadSpec` failure, any idea what that might be about? I don't have a record of the full error, let's ask Jenkins for one: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1146/consoleFull (link will be 404 for a while the poor overworked chap attends to some other things first)

SethTisue:
@rossabaker hmm, that test run failed with `[http4s] [error] Failed tests:
[http4s] [error] 	org.http4s.json4s.jackson.Json4sJacksonSpec` instead

rossabaker:
Neat, I didn't know we were being included!  

Fair warning: blaze-client and async-http-client ~~and server~~ are all not compiling at the moment on that branch.  I'm surprised these didn't show, but I admit to not having done my homework yet on dbuild.  If scalaz-stream could be added, master would be a more stable choice.  If not, maybe should be excluded until topics/cats is stabilized, which I hope is very soon.

As for the failure, that module passes on topics/cats.  It's possible that there's a breaking change in the latest json4s-jackson.  I'll see if I can test that later.

sarahgerweck:
@SethTisue: merged your Log4s PR. If there's anything else that will make this easier on you, feel free to ask (or submit a PR!).

SethTisue:
> blaze-client and async-http-client and server are all not compiling at the moment on that branch. I'm surprised these didn't show

@rossabaker if you look closely at the diffs in this PR, you'll see that I excluded those subprojects. we _could_ re-add scalaz-stream to the community build (we used to have it), but I'm happy to just wait  for now and hope that `topic/cats` gets merged (if you think of it, you could ping me when it does).

@sarahgerweck thanks!

PR #467, opened 2017-02-04
fix play-ws

SethTisue:


SethTisue:
fyi @wsargent

already tested locally.

wsargent:
Thx

PR #479, opened 2017-02-16
fix argonaut and upickle-pprint

SethTisue:
just the usual sorts of issues with scalaVersion and
unmanagedSourceDirectories


SethTisue:
tested locally

PR #486, opened 2017-03-01
remove scalafmt workarounds

SethTisue:
latest scalafmt versions are more dbuild-friendly.

scala-json-ast already upgraded scalafmt themselves.

akka-sse was still on an old version.  we are already forking akka-sse,
so I refreshed the fork and added a new commit with the scalafmt upgrade.
I also separately submitted that commit as a PR.  (even if it's accepted,
we'll still need to stay forked, because sbt-bintray.)

### DIFF
--- a/community.dbuild
+++ b/community.dbuild
@@ -767,13 +767,6 @@ build += {
   ${vars.base} {
     name: "scala-json-ast"
     uri:  ${vars.uris.scala-json-ast-uri}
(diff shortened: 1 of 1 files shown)


SethTisue:
already tested locally

SethTisue:
after Heiko merged https://github.com/hseeberger/akka-sse/pull/172, I rebased the branch in our fork onto latest master

PR #487, opened 2017-03-02
unfreeze utest

SethTisue:
(Lukas had frozen it in June 2016, not sure why, presumably some thing
involving utest not having 2.12 support fully in place yet, since
addressed)


SethTisue:
tested locally.

this is nice to have unfrozen in general, but my specific motivation is that I want it unfrozen in the 2.13 community build, where it's failing, and I want to submit an upstream PR that will fix it, so we need to be unfrozen to pick that up once it gets merged

PR #496, opened 2017-03-15
add typesafe-emoji

SethTisue:
we cannot risk regressions in this critical piece of
infrastructure.  we will never let our customers down.


SethTisue:
tested locally

dwijnand:
I've already picked the name for my island in the Caribbean, have you picked yours?

SethTisue:
the islands we'll flee to when the villagers come after us with torches and pitchforks?

PR #511, opened 2017-04-04
re-add scala-gopher (to 2.12)

SethTisue:
it is in the 2.11 build. I couldn't find the commit where it got
dropped here, perhaps it happened inadvertently?


SethTisue:
tested locally

SethTisue:
fyi @rssh

PR #514, opened 2017-04-06
unfreeze scalaz

SethTisue:
thanks @xuwei-k for backporting the needed Scala Native workaround
from 7.3.x to 7.2.x


SethTisue:
tested locally

PR #521, opened 2017-04-10
disable fatal warnings for a few projects

SethTisue:
because of scala/scala#5402

fyi @fommil @tpolecat

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -588,9 +588,12 @@ build += {
     extra.run-tests: false
   }
 
+  // disable -Xfatal-warnings because of new unused warnings introduced
+  // by https://github.com/scala/scala/pull/5402
   ${vars.base} {
     name: "pcplod"
     uri:  ${vars.uris.pcplod-uri}
+    extra.commands: ${vars.default-commands} [ "removeScalacOptions -Xfatal-warnings" ]
   }
 
   ${vars.base} {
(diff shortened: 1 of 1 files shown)


configs/community.dbuild
```diff
@@ -588,9 +588,12 @@ build += {
     extra.run-tests: false
   }
 
+  // disable -Xfatal-warnings because of new unused warnings introduced
+  // by https://github.com/scala/scala/pull/5402
   ${vars.base} {
     name: "pcplod"
     uri:  ${vars.uris.pcplod-uri}
+    extra.commands: ${vars.default-commands} [ "removeScalacOptions -Xfatal-warnings" ]
```
adriaanm, line 596: is the error in pcplod going to be fixed by this? I think it's an error in its test suite, which is likely not affected by the compiler used to compile the project?

adriaanm:
@SethTisue, Did you also mean to also do this for kind-projector? See also my comment about pcplod

SethTisue:
@som-snytt's [kind-projector PR](https://github.com/non/kind-projector/pull/48) was already merged

as for pcplod, I'll see how the run goes and take appropriate action.

adriaanm:
ah, 👍

adriaanm:
I pro-actively filed https://github.com/ensime/pcplod/issues/20

configs/community.dbuild
fommil: interesting, I should change the `scalacOptions` used in test to be definitive and not inherit from the build.

fommil: I'm travelling right now, so might be the weekend before I get to this.

PR #576, opened 2017-07-21
add better-files

SethTisue:
fyi @pathikrit

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -139,6 +139,54 @@ build += {
   projects: [
 
   ${vars.base} {
+    name: "scala-parser-combinators"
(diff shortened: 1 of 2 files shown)


SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1685/consoleFull ~~(404 while Jenkins ponders that there is iced tea _inside the handle_ of the iced tea jug even though the handle _is itself the thing_ used to lift the tea)~~

SethTisue:
note that for now we're just compiling the test suite, not running it, because of https://github.com/pathikrit/better-files/issues/165

PR #639, opened 2017-11-25
use cloc to count lines of code compiled

### LINKED ISSUE #59: Track automatically how many lines of code we are compiling
We are tracking number of lines compiled by community build by manually running `cloc` from time to time. It would be great to run cloc automatically.

See also typesafehub/dbuild#133 and typesafehub/dbuild#142

SethTisue:
fixes #59

fork sbinary because there were two different ways it was calling
`update` during dependency extraction, which fails since cloc-plugin
isn't published anywhere

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -80,6 +80,11 @@ set commands ++= {
 vars.default-commands += "appendScalacOptions "${vars.scalac-opts}
 vars.base.extra.commands = ${vars.default-commands}
 
+//// count lines of code
+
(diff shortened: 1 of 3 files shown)


SethTisue:
I'm going to merge despite not having a fully green test run in hand — I've seen it green on Jenkins with a reduced projects list, and a full run (https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2281/console) seems to be doing well so far — if a project or two still fails post-merge, I'll deal with it then.

PR #676, opened 2018-01-29
build Scala.js stubs project separately

SethTisue:


SethTisue:
test run https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2588/consoleFull

SethTisue:
the test run is still in progress, but I can already see that in dbuild's view of what depends on what, all of the dependencies on scala-js are now on scala-js-stubs instead, as intended

SethTisue:
and in cfc6eff0386c9132d3a8ae073895eb46d81ac530 I froze scala-js-stubs, since it hardly ever changes, so we want to avoid having to rebuild downstream projects just because something changed elsewhere in the repo (which happens _all the damn time_, because @sjrd and crew are so industrious)

PR #618, opened 2017-10-26
add sttp and fs2-reactive-streams

SethTisue:
fyi @adamw

PR #678, opened 2018-02-01
unfork sbinary

SethTisue:
in a local test, sbinary couldn't find the cloc-plugin dependency.
no idea what that's about.

I should return to it once I clear my plate of some higher-priority
tasks.

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -458,18 +458,17 @@ build += {
     ]
   }
 
-  // forked (October 2017) to work around unfriendliness between dbuild
-  // and sbt-houserules 0.3.4; details at
-  // https://github.com/scala/community-builds/issues/603 . we should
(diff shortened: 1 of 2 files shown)


SethTisue:
> sbinary couldn't find the cloc-plugin dependency

now it's happening over at #618 too

SethTisue:
oh, this is familiar. the cause is that `update` is being called during dependency extraction. usually Scalafmt related. in sbt-io it was sbt-houserules related.

SethTisue:
perhaps relevant: lucidsoftware/neo-sbt-scalafmt@545079f93f3fdf428c9611b7eb7f575bf978cf59 (note the use of `update.value`)

SethTisue:
I think this one just needs the same workaround we used for sbt-io (and elsewhere). but the sttp one seems harder

SethTisue:
(green in local test)

PR #684, opened 2018-02-12
reinstate some projects to the 2.13 build

SethTisue:
usually because sbt 1 supports 2.13 now


SethTisue:
green in local testing

fyi @milessabin (Shapeless re-added)
fyi @wsargent (lightbend-emoji triumphantly and gloriously reinstated)

PR #688, opened 2018-03-04
Adds the singleton-ops library to the community build

soronpo:


SethTisue:
the repo readme seems to say that Typelevel Scala is required. is that not the case...?

SethTisue:
perhaps we should only attempt to add this on the 2.13.x branch, where SIP-23 has been merged? but the repo doesn't have seem to have 2.13 support.

regardless, I'll do a test run and see what happens.

soronpo:
> the repo readme seems to say that Typelevel Scala is required. is that not the case...?

No. singleton-ops works with LBS, as well. Its best use-case is SIP-23, but one can use it, just like people manage to use `refined` with shapeless Witness.

SethTisue:
green in local test after fixups in e3b8257c0651d129f6fcb8a26935829e3d818fa2. thanks Oron!

SethTisue:
I've merged this forward onto 2.13.x, where it will fail for a while because the latest 2.13.x nightly is weeks old because of recent Jenkins upheavals

PR #692, opened 2018-03-16
Add metrics-scala.

erikvanoosten:
As far as I can see these is nothing special about metrics-scala's build so I've put nothing in `community.dbuild`.


SethTisue:
in your build.sbt can you update this logic

```
      case v if v.startsWith("2.12") => Tests.Argument("-l", "<scala2.12")
```

to accommodate Scala 2.13 as well? since this will be merged forward into the 2.13 community build as well

SethTisue:
let's continue on #693

PR #703, opened 2018-03-27
add refined

SethTisue:
/cc @fthomas @soronpo @fommil

Oron reminded me this was still missing.  I had tried to add it a year
ago but failed, I don't remember why. let's see how it goes this time.

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1779,6 +1779,15 @@ build += {
     ]
   }
 
+  // in this space because it depends on pureconfig
+  ${vars.base} {
+    name: "refined"
+    uri:  ${vars.uris.refined-uri}
+    extra.sbt-version: ${vars.sbt-1-version}
+    // I don't see a project that aggregates just the JVM subprojects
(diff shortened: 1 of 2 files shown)


fthomas:
Many thanks, @SethTisue!

Are sbt subprojects that aggregate all JVM subprojects common? If you think this is beneficial for the community build, we can add such project to the refined build.

SethTisue:
@fthomas it's pretty common. typical naming would be `refinedJVM`. it's a fairly minor issue, but, it would ensure that if you add a new subproject, it will get picked up here.

it can also be useful for contributors who want to get all the JVM stuff working and then worry about Scala.js and/or Scala Native in a separate step. (it is fairly common for prospective open source contributors not to have familiarity or experience with Scala.js or Scala Native.)

fthomas:
Ok, that makes sense. I've opened https://github.com/fthomas/refined/issues/460 to track that in refined. Thanks again @SethTisue.

fommil:
you can also use https://github.com/sjrd/sbt-dynscalajs which makes the JVM the first class citizen.

Or, like me, you can just let the scalajs community publish independent artefacts of your projects or have source dependencies. Cross building for js and native has become such a huge burden for library authors.

PR #708, opened 2018-04-04
Add typelevel/mouse to build

benhutchison:
Full disclosure: after a week of (re-)trying I havent been able to get the build to run green locally on my laptop at home. Problem seems to be that after 1-3 hours of running transient glitches in my home internet cause the build to fail, eg `fatal: unable to access 'https://github.com/sksamuel/exts.git/': Could not resolve host: github.com`

A quicker "sanity check" build would be useful for checking PRs

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -658,6 +658,7 @@ build += {
   ${vars.base} {
     name: "mouse"
     uri:  ${vars.uris.mouse-uri}
+    extra.sbt-version: ${vars.sbt-1-version}
     extra.projects: ["crossJVM"]  // no Scala.js please
   }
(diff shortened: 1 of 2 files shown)


SethTisue:
I don't see any way to do a sanity check. a good fix would be https://github.com/scala/scala-jenkins-infra/issues/92

SethTisue:
the local-runs experience is a lot smoother if you run Artifactory locally (there is a config file in this repo for this, which I use), but it's not worth the hassle of setting up unless you do community build work on a fairly regular basis.

SethTisue:
thanks! glad to have this. green in local testing.

PR #709, opened 2018-04-05
Add Eff to build

benhutchison:


SethTisue:
dbuild isn't able to resolve the scalameter dependency in coreJVM. I don't know why that's happening, but I'd rather not have to troubleshoot it. in general, we prefer to exclude benchmarks from the community build. could you move the scalameter stuff to a separate subproject, which we can then exclude?

benhutchison:
Seth, Ive moved the benchmark out of coreJVM in https://github.com/atnos-org/eff/pull/136

SethTisue:
thanks!

next issue:

```
[eff] [error] /Users/tisue/community.212/target-0.9.11/project-builds/eff-cb5fd474298c93c4fbaa0be2a22510b29ab7592f/shared/src/main/scala/org/atnos/eff/SafeEffect.scala:217: method apply in object PartialFunction is deprecated (since 2.12.5): For converting an ordinary function f to a partial function pf, use `val pf: PartialFunction[A, B] = { case x => f(x) }`. For creating a new PartialFunction, use an explicit type annotation instead, like in `val pf: PartialFunction[Int, String] = { case 1 => "one" }`.
[eff] [error]     recoverThrowable[R, A, B](action, pureValue, PartialFunction(onThrowable))
[eff] [error]                                                  ^
[eff] [error] one error found
```

I could work around this by disabling `-Xfatal-warnings` in the context of the community build, but I figure you'll probably want to address it directly.

benhutchison:
Thanks Seth, that’s fixed now.

Ben

SethTisue:
```
[info] Project eff-------------------------: SUCCESS (project rebuilt ok)
```

I'll merge after we unfreeze the build post-2.12.6

SethTisue:
thanks, glad to have this.

PR #711, opened 2018-04-17
add Magnolia

SethTisue:
but note that we don't run the tests because there's a shell
script involved, rather not mess with it if `sbt test` doesn't
do it


SethTisue:
fyi @propensive

SethTisue:
in local testing:

```
[info] Project magnolia--------------------: SUCCESS (project rebuilt ok)
```

PR #713, opened 2018-04-26
make scala-xml a regular project

SethTisue:
it doesn't need to be special anymore now that it's out of the bootstrap

fyi @ashawley


SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/1051/consoleFull (404 for now)

SethTisue:
test run: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/1053/consoleFull

SethTisue:
okay, we don't need to wait for the whole thing to run, I've seen enough to be sure this is mergeable

SethTisue:
ahhhhhhhhhhhhhhhhhhhhhh

SethTisue:
gah, in the commit message for b31ba8e I meant to say "assemble" build system, not "scala" build system

and also b31ba8e is actually wrong, because it no longer sets `scalaVersion` for every included project. I need to add something back in that does that. working on it.

cunei:
@SethTisue I think you will still need to fetch via aether scala-library etc, using the assembly build system. Compare the dependency information section before:
```
[info] Project sbinary
[info]   depends on: cloc-plugin, scala
[info] Project scala-partest
[info]   depends on: cloc-plugin, scala
[info] Project sbt-testng
[info]   depends on: cloc-plugin, scala
[info] Project kind-projector
[info]   depends on: cloc-plugin, scala
```
and after:
```
[info] Project sbinary
[info]   depends on: cloc-plugin, scala-xml
[info] Project scalatest
[info]   depends on: cloc-plugin, scala-xml
[info] Project scala-records
[info]   depends on: cloc-plugin, scala-xml, scalatest
```
There is no longer a reference to a project that contains scala-library and scala-compiler. If dbuild cannot find a custom scala-library and scala-compiler as part of the build, it cannot adjust `scalaVersion` and `scalaHome` in order to refer to them, and it will just use the scala version defined by each project. So, you are not really testing the 2.13.x head anymore.

SethTisue:
@cunei right, I'd figured that out, so the fix I'm trying is to set the Scala version in vars.base.extra.commands. you think it's better to just bring back assemble?

SethTisue:
I've reverted b31ba8e27da9133cc72ddd81a485ce55ea86db63

PR #715, opened 2018-05-01
specs2 3.x -> 4.x

SethTisue:
the 2.13 branch is already on 4.x, let's try 2.12

it's generally good to be on newer stuff, but my specific motivation
here for doing this on 2.12 is to get better feedback when we run the
2.12 community build on pull requests.  specs2 4.x drops the scalaz
dependency, so that gives us a flatter dependency tree.  scalaz is
pretty tricky code, so it's common that it breaks, knocking out specs2
3.x, knocking many downstream projects out of the build.

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -384,7 +384,7 @@ build += {
     ]
   }
 
-  // this is separate from "akka" because there is a circular dependency between
+  // this is separate from "akka-actor" because there is a circular dependency between
   // the akka and ssl-config repos
   ${vars.base} {
     name: "akka-more"
@@ -967,6 +967,8 @@ build += {
     extra.projects: ["minitestJVM", "lawsJVM"]  // no Scala.js
   }
 
+  // frozen (May 2018) at an April 2018 commit just before an upgrade to a
+  // cats-effect version that isn't source compatible with the one we have
   ${vars.base} {
     name: "monix"
     uri:  ${vars.uris.monix-uri}
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -678,26 +678,39 @@ build += {
     extra.projects: ["crossJVM"]  // no Scala.js please
   }
 
-  // forked (January 2018) to remove a couple of failing tests; ticket
-  // is https://github.com/etorreborre/specs2/issues/622. note we are
-  // using the 3.x branch, even though specs2 4.0 has been out for a while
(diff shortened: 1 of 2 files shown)


SethTisue:
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2791/ queued (but 404 for a while yet, the queue is pretty backed up right now)

SethTisue:
> specs2 4.x drops the scalaz
> dependency

oh, well, it's still there but only in a special `scalazJVM` subproject. so I'll need to either omit that subproject, or build it in a separate entry in the config file.

EDIT: oh, it might be used downstream, so I'll go the separate-entry route.

SethTisue:
in local testing, I've already verified that the benefit will be substantial if we can get this through:

before this change, 79 projects depended on scalaz (that's half the build!)
after, only 20

SethTisue:
okay the separated specs2-scalaz entry is now green in local testing... we'll see what fails downstream

SethTisue:
tons of green projects in a partial run locally, so this looks really promising. Jenkins run in progress now: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2796/consoleFull

SethTisue:
~~https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2797/consoleFull~~
~~https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2799/consoleFull~~
~~https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2801/consoleFull~~
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2802/consoleFull

SethTisue:
using the `SPECS2-4.0.3` tag to avoid https://github.com/playframework/playframework/issues/8393 didn't work; specs2 itself had a test failure. rather than report that test failure, which is a bit of a waste since it's only in an old version, let's just disable that part of play's tests for now; there's a good chance the Play team will fix the problem at their end now that I've reported it (it's easy to reproduce outside of dbuild)

SethTisue:
this was much easier than I feared.

fyi @etorreborre

etorreborre:
You don't have an easy job :-).

Issue #609, opened 2017-09-24
add a Java 9 community build


SethTisue:
starting, I suppose, with just a stripped-down thing that only attempts to build some basic stuff like the Scala standard modules, then expanding from there

Scala 2.12 only is fine for now, I think. the behemoths are already somewhat pressed for space (and the space can't be increased without rebuilding the nodes), and even just adding the 2.12/jdk9 combo will make the problem, one might say, worse enough

plan:
* get it running locally on my Mac
* get it running on Jenkins

re: Jenkins, @retronym did some work, prior to JDK 9 going final/gold/GA, on making it available on our Jenkins nodes; see https://github.com/scala/scala-jenkins-infra/issues/174 for details

SethTisue:
my experience attempting this locally was that all extractions failed with `[scalacheck] [error] scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found.`

but then I made an empty sbt project and ran it with Java 9, and https://github.com/sbt/sbt-launcher-package/pull/143 kicked in and showed the "Copying runtime jar." message. I thought then running dbuild again might work, but same error

maybe I also need to do something to make sure `-Dscala.ext.dirs=...` gets set?

SethTisue:
> maybe I also need to do something to make sure -Dscala.ext.dirs=... gets set?

helps, but is not sufficient. adding `extra.options: ["-Dscala.ext.dirs=/Users/tisue/.sbt/0.13/java9-rt-ext-oracle_corporation_9"]` to every projects allows extraction to proceed farther, but then eventually every projects gets an error like:

```text
[scala-parser-combinators] java.lang.AssertionError: assertion failed: not a type: symbol
[scala-parser-combinators]  loaded from class ScalaModulePlugin$ in package with name ScalaModulePlugin$ and classloader sbt.PluginManagement$PluginClassLoader@9b8d3db
[scala-parser-combinators] 	at scala.reflect.runtime.JavaMirrors$JavaMirror.scala$reflect$runtime$JavaMirrors$JavaMirror$$classToScala1(JavaMirrors.scala:979)
[scala-parser-combinators] 	at scala.reflect.runtime.JavaMirrors$JavaMirror$$anonfun$classToScala$1.apply(JavaMirrors.scala:938)
[scala-parser-combinators] 	at scala.reflect.runtime.JavaMirrors$JavaMirror$$anonfun$classToScala$1.apply(JavaMirrors.scala:938)
```

[...]

PR #725, opened 2018-05-12
drop scala-debugger

SethTisue:
because its build definition doesn't compile on JDK 9, and we don't want
to maintain separate JDK 8 and 9 config files

reference: https://github.com/scala/community-builds/issues/609

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1292,7 +1292,9 @@ build += {
   }
 
   // dependency of scala-debugger. we track the develop branch, since
-  // master seems neglected/outdated
+  // master seems neglected/outdated.
+  // note that scala-debugger has been pulled from the build (perhaps
+  // not permanently), so if this causes trouble, we might drop it too.
   ${vars.base} {
     name: "scallop"
     uri:  ${vars.uris.scallop-uri}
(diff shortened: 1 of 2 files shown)


SethTisue:
@chipsenkbeil if you're interested in helping get it back in, let me know. the minimum requirement is that the build definition needs to compile on JDK 9. the project doesn't need to actually work on JDK 9, that can be postponed.

currently what we're hitting is:

```
[info] Loading project definition from /Users/tisue/scala-debugger/project
java.io.FileNotFoundException: 
Could not automatically find the JDK/lib/tools.jar.
You must explicitly set JDK_HOME or JAVA_HOME.
    
	at org.scaladebugger.SbtJdiTools$$anonfun$6.apply(SbtJdiTools.scala:45)
	at org.scaladebugger.SbtJdiTools$$anonfun$6.apply(SbtJdiTools.scala:45)
	at scala.Option.getOrElse(Option.scala:120)
	at org.scaladebugger.SbtJdiTools$.<init>(SbtJdiTools.scala:44)
	at org.scaladebugger.SbtJdiTools$.<clinit>(SbtJdiTools.scala)
	at java.base/java.lang.Class.forName0(Native Method)
```

which stems from this plugin:

```
addSbtPlugin("org.scala-debugger" % "sbt-jdi-tools" % "1.0.0")
```

which would need to be updated to work, or at least not fail hard, on JDK 9

chipsenkbeil:
@SethTisue, looks like tools.jar was removed and the relevant files are loaded during bootstrap if using the JDK from JDK 9: https://issues.apache.org/jira/browse/MNG-5732

I wonder if adding a condition to only load the sbt-jdi-tools plugin if the JDK is less than 9 would solve the problem.

chipsenkbeil:
@SethTisue, newest commit to scala debugger project enables you to load the project with JDK 9. It doesn't compile yet on JDK 9, I think because I'm using an old version of sbt which uses Scala 2.10.

SethTisue:
@chipsenkbeil build definition loads now, thanks! removal reverted in d6c484766fe843ecd57a521ade5861177000a41d

PR #739, opened 2018-06-27
adjust Lagom tests

SethTisue:
there is a new recurring failure we need to address

but also, now that I know how to exclude failing tests per-source-file rather than the coarser per-subproject, let's re-enable tests in the subprojects we'd excluded and see what fails


SethTisue:
test run queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3057/ (404 for now)

SethTisue:
green at https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3068/ (scalameter failure is unrelated)

PR #752, opened 2018-07-18
unfreeze scalameta (to get scalafix working again)

SethTisue:
because scalafix master now needs scalameta's new symtab module

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1686,9 +1686,6 @@ build += {
     check-missing: false
   }
 
-  // frozen at version that currently works for both scalafix and scalafmt.
-  // (tracking master of all three projects is too fragile).
-  // current tag chosen to include necessary geny fix, JDK 9 fix
   ${vars.base} {
     name: "scalameta"
     uri:  ${vars.uris.scalameta-uri}
@@ -1699,7 +1696,7 @@ build += {
       // to include.  this is the main thing:
       "semanticdbScalacPlugin"
       // and I empirically found scalafix and/or scalafmt to need:
-      "testkit", "contribJVM"
+      "testkit", "contribJVM", "symtab"
     ]
     // use right version-specific source directories regardless of our weird dbuild Scala version numbers
     extra.commands: ${vars.default-commands} [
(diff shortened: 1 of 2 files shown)


SethTisue:
test run: 
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3173/

SethTisue:
if this doesn't work out, we should just freeze scalafix at the last green commit, that would be perfectly fine too /cc @olafurpg

olafurpg:
FWIW I cleaned up the scalafix build yesterday, upgraded to sbt 1, removed Scala.js, and generally simplified the build a lot. Scalafix should work with the latest scalameta master.

SethTisue:
okay, the question then is whether something else will break :-)

SethTisue:
queued: 
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3175/ (404 for now)

SethTisue:
@olafurpg I had to add `unit` to `extra.excludes`, because it was trying to compile stuff from `testsOutputDotty`. but as a result we don't seem to be running any scalafix tests at all.

offhand, it looks like it might be because:

```
    compileInputs.in(Compile, compile) := {
      compileInputs
        .in(Compile, compile)
        .dependsOn(
          compile.in(testsInput, Compile),
          compile.in(testsOutputDotty, Compile),
```

and as for why it would fail in the community build only, I imagine it's because your build is setting `scalaVersion` to (e.g.) 0.9.0-RC1 and then dbuild overrides that to whatever Scala version the community build is using

can the Dotty stuff be kept more separate?

olafurpg:
@SethTisue My goal is to move the dotty rules into a separate repository soon similar to scala-collection-compat. First I need to improve the support for externally published rules, which is in my queue before scalafix v0.6.

olafurpg:
The dotty rules have now been moved out of the scalafix build so that sbt-dotty should no longer cause problems.

SethTisue:
@olafurpg awesome! but I think unfreezing would still run afoul of #765, let's continue there

PR #765, opened 2018-08-10
newer coursier, scalafix, scalameta; add coursier-small; pull ammonite

SethTisue:
because scalafix now depends on coursier-small

/cc @olafurpg

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -2095,9 +2095,6 @@ build += {
     ]
   }
 
-  // see also ammonite-ops.
-  // frozen (April 2018) at a March 2018 commit before a problematic upickle
-  // upgrade; see also upickle
   ${vars.base} {
     name: "ammonite"
     uri:  ${vars.uris.ammonite-uri}
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1839,13 +1839,7 @@ build += {
     name: "scalafix"
     uri:  ${vars.uris.scalafix-uri}
     extra.sbt-version: ${vars.sbt-1-version}
-    extra.exclude: [
-      "website", "testsOutputDotty"
-      // this seems to somehow pull in testsOutputDotty unless we exclude it, I don't understand why.
-      // but by excluding it, we aren't running any scalafix tests at all; see
-      // https://github.com/scala/community-builds/pull/752
-      "unit"
-    ]
+    extra.exclude: ["website", "docs"]
   }
 
   ${vars.base} {
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1764,7 +1764,7 @@ build += {
       // to include.  this is the main thing:
       "semanticdbScalacPlugin"
       // and I empirically found scalafix and/or scalafmt to need:
-      "testkit", "contribJVM", "symtab"
+      "testkit", "contribJVM", "symtab", "metapJVM"
     ]
     // use right version-specific source directories regardless of our weird dbuild Scala version numbers
     extra.commands: ${vars.default-commands} [
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -2087,6 +2087,22 @@ build += {
     ]
   }
 
+  ${vars.base} {
+    name: "scala-debugger"
+    uri:  ${vars.uris.scala-debugger-uri}
+    extra.exclude: [
+      // no sbt plugins plz!
+      "sbtScalaDebuggerPlugin"
+      // Missing dependency: the library org.senkbeil#grus-layouts
+      "scalaDebuggerDocs"
+      // depends on ammonite, which we no longer have
+      "scalaDebuggerTool"
+    ]
+    // java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: GC overhead limit exceeded
+    // perhaps only the -Xmx is really necessary, but let's try it with all of the repo's own .jvmopts:
+    extra.options: ["-Xms1g", "-Xmx4g", "-Xss2m", "-XX:MaxMetaspaceSize=256m"]
+  }
+
 ]}
 
 //// space: jawn_0_11
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1853,6 +1853,13 @@ build += {
     uri:  ${vars.uris.scalafix-uri}
     extra.sbt-version: ${vars.sbt-1-version}
     extra.exclude: ["docs"]
+    extra.commands: ${vars.default-commands} [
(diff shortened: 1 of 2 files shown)


SethTisue:
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3288/ (404 til Jenkins queue clears)

olafurpg:
Thanks, I didn't think of this when I added the dependency 😅

SethTisue:
mercator is green (and so is magnolia)

`coursier-small`:

```
[coursier-small] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.14/project-builds/coursier-small-ef765717232c2339bf41aced25e10801fd2dea56/small/src/main/scala/com/geirsson/coursiersmall/CoursierSmall.scala:5:8: object Gather is not a member of package coursier.util
[coursier-small] [error] import coursier.util.Gather
[coursier-small] [error]        ^
```

`Gather` wasn't added to coursier until a bit later than the commit we're frozen at.

unfreezing coursier might break ammonite, but that's something we'd probably need to deal with sooner or later regardless

SethTisue:
I moved the mercator/magnolia part to #768

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3291/ (404 til Jenkins queue clears)

SethTisue:
> unfreezing coursier might break ammonite, but that's something we'd probably need to deal with sooner or later regardless

indeed, it did. (the other failures in run 3291 are unrelated)

I don't have time to deal with it until after 2.13.0-M5 is out, so for now I'll just freeze scalafix at the last green commit, before the coursier-small addition

not sure how hard moving ammonite to the new version might be.

one possibility might be to introduce a space split and have both old-coursier and new-coursier in the build

olafurpg:
I can probably replace the usage of coursier-small in the scalafix repo with https://github.com/sbt/librarymanagement if that helps. It's only used for a single integration test that can run in 2.12 only. I only strictly need coursier-small in sbt-scalafix to support 0.13.

olafurpg:
Or even easier, I can downgrade the coursier dependency version in coursier-small to match the ammonite version. 

I'd prefer not to change things right away, hopefully it can be sorted out. But if this continues to be a problem I'm happy to accommodate.

SethTisue:
let's get a fresh read on the current status: 
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3385/

SethTisue:
try again: 
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3395/ (404 for now)

SethTisue:
```
[scalafix] [error] **** Missing dependency: the library org.scalameta#metap is not provided (in space "scala.main") by any project in this configuration file.
```

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3403/

SethTisue:
`failed: scalafix` because of compile errors, apparently we need a different scalameta version

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3411/

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3414/

SethTisue:
okay, run 3414 has scalafix green, but we still need to see what else breaks: 
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3416/

SethTisue:
as expected, ammonite breaks

but also, classpath-shrinker breaks, I think because the even-newer coursier doesn't have a scalaz dependency anymore

SethTisue:
https://github.com/scalacommunitybuild/classpath-shrinker/commit/ac7cff06f4f9e5b157a9da78ea34974acfe57d44 should take care of the classpath-shrinker issue

SethTisue:
the only project downstream from ammonite proper (as opposed to ammonite-ops) is scala-debugger. if everything else is green, I'm rather tempted to just pull ammonite and scala-debugger from the build

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/3423/

SethTisue:
oh, you know what, we have Ammonite frozen at an old version, on master it's already on coursier 1.1.0-M7, which is where we want to be

so let's unfreeze Ammonite and see what happens

SethTisue:
gah, Ammonite isn't even built with sbt anymore

perhaps eventually we'll need to add mill support to dbuild, but that isn't happening anytime soon, so I guess I'll just pull ammonite and scala-debugger for now

SethTisue:
> guess I'll just pull ammonite and scala-debugger for now

oh, we should only have to pull the one subproject of scala-debugger that depends on ammonite, not pull the whole thing

SethTisue:
fyi @chipsenkbeil @lihaoyi

@olafurpg there's nothing you need to do here, unless you have some insight into this:

```
      // java.lang.IllegalStateException: unable to auto-detect semanticdb-scalac compiler plugin
       //   at scalafix.tests.cli.ScalafixImplSuite.$anonfun$semanticdbPluginPath$1(ScalafixImplSuite.scala:37)
       //   at scala.Option.getOrElse(Option.scala:121)
       //   at scalafix.tests.cli.ScalafixImplSuite.semanticdbPluginPath(ScalafixImplSuite.scala:36)
       "set excludeFilter in (Test, unmanagedSources) in unit := HiddenFileFilter || \"ScalafixImplSuite.scala\""
```

chipsenkbeil:
@SethTisue, ammonite is used for the __scala-debugger-tool__ and subsequent __sbt-scala-debugger-plugin__ modules. Could you just exclude those? In fact, I think you're already excluding the sbt plugin, right? The main module that serve as good build test is the api one.

olafurpg:
@SethTisue Thanks! The test failure is not surprising if the classloader's classpath is different than from when opening sbt normally. I generally try to avoid writing such tests but this was the quickest way to get a full e2e test, it's fine to disable the test.

When I think of it, if you disable `ToolClasspathSuite` as well then you can remove the coursier-small dependency in the scalafix build.

SethTisue:
> @SethTisue, ammonite is used for the scala-debugger-tool and subsequent sbt-scala-debugger-plugin modules. Could you just exclude those?

@chipsenkbeil yes, that's what I ended up doing.

@olafurpg the coursier-small dependency isn't really causing any trouble, but if it does, I'll keep it in mind that it would be easy to remove.

olafurpg:
@SethTisue Excluding `ToolClasspathSuite` on your side should make the dependency unnecessary

PR #769, opened 2018-08-13
add my Project Euler solutions

SethTisue:


SethTisue:
not to be merged yet, because it turns out that my solution to problem 95 finishes in 6 seconds on 2.12.6 but hangs on 2.12.7-bin-4779c42 so that's a likely Scala regression

SethTisue:
problem 75 hangs also. the rest are okay

SethTisue:
I've tweaked this to exclude problems 75 & 95 for now, until https://github.com/scala/scala/pull/7091 is merged

green in local testing

PR #785, opened 2018-09-06
unfreeze jackson-module-scala (and enable some disabled things downstream, and pull scalameter)

SethTisue:
this isn't a totally crucial upgrade, but it's desirable

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -837,6 +837,12 @@ build += {
       // the `-bootclasspath ...` thing their build does doesn't work on Java 9
       "set javacOptions ~= (_.dropRight(2))"
     ]
+    // without this, in downstream projects we get:
+    // com.fasterxml.jackson.databind.JsonMappingException:
+    //   Scala module 2.8.8-dbuildx1c5888e4b2a136333f7edb02dbe6dc5abbfa7de1 requires Jackson Databind version >= 2.8.0 and < 2.9.0
+    // we could submit an upstream PR adjusting the regex that's resulting in this
+    // error, but it's easier to just:
+    set-version: "2.8.7"
   }
 
   // tracking "develop" branch.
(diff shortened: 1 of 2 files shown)


SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3452/

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3455/

SethTisue:
haha oops FAIL

```
[twitter-util] [info]   com.fasterxml.jackson.databind.JsonMappingException: Scala module 2.8.7 requires Jackson Databind version >= 2.8.0 and < 2.9.0
```

I changed the version of the wrong thing, apparently

SethTisue:
so as we see from https://github.com/FasterXML/jackson-module-scala/issues/265 , the error message annoyingly doesn't include the offending version number, the one returned by `context.getMapperVersion`, so we don't even know what wrong version the check thinks it's finding

SethTisue:
okay so what I thought was going on here, wasn't right at all.

I think I may know what's going on here. we have jackson-module-scala frozen at an old version. downstream projects such as twitter-util and akka-http-json may be declaring a dependency on a newer version. because dbuild rewires Scala-based dependencies, as long as everything's source compatible it's fine.... or would it be, if it weren't for this version check, which is a check on the version number of a Java-based dependency, not a Scala-based one. that Java-based dependency remains un-rewired and thus won't match the version our frozen jackson-module-scala expects, and boom.

I'd have to actually verify this theory, but I bet it's right.

our config file says:

```
  // frozen (July 2018) at a February 2017 commit, because there were
  // scalameter test failures when we used master.  (whether going all the
  // way back to February 2017 is actually necessary, I don't know; I
  // just used the last green commit, which happens to be old because
  // we had the project frozen already for a long time)
```

scalameter is a leaf project, so it shouldn't be the driver on this.

twitter-util is on 2.9.6, and so is akka-http-json!

so we should move to that version, and if scalameter breaks as a result, cross that bridge when we come to it.

SethTisue:
asked scalameter if they are willing to upgrade: https://github.com/scalameter/scalameter/issues/205

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3458/consoleFull

SethTisue:
compile errors in the jackson module of akka-http-json because it doesn't declare its scala-reflect dependency, and I guess isn't getting it transitively anymore in the versions of things we have here

well, it's easy to inject the dependency

Issue #798, opened 2018-10-03
possible to include at least a little Scala.js stuff?


SethTisue:
as I wrote at https://github.com/scala/bug/issues/11174 :

>> Actually is it still the case that you can't use Scala.JS in the community build?

> We've historically been reluctant to expand the scope of the community build to include the Scala.js ecosystem, not primarily for technical reasons, but primarily because the community build already consumes a lot of engineering resources so we have to be skeptical of any further expansions of scope. But perhaps we could at least include a handful of things?

/cc @sjrd @japgolly

nafg:
Isn't it well worth it? Scala.js is a major part of the scala ecosystem nowadays.

ngbinh:
Please do. Scala.js and scala.js react plays a vital role in our tech stack.

SethTisue:
first stab at #799 

@nafg time and money are finite, these "worth it" judgements are extremely hard to make. it is clearly desirable, yes, there's no question. but for every thing we take on at Lightbend or the Scala Center or LAMP, there's 10 or 100 other things we can't take on that are also desirable

SethTisue:
P.S. note that we have enough capacity on our Jenkins, that isn't the issue. so if somebody in the community wants to put together a Scala.js community build config file, to be community-maintained going forward, I'd be happy to make a Jenkins job for it, assuming the dbuild issue we're hitting at #799 can be resolved (which is unclear)

ngbinh:
@SethTisue are there any guide on how to help to add more projects to the community build? I would love to help if I can

SethTisue:
@ngbinh yup, lots of docs in https://github.com/scala/community-builds/wiki

SethTisue:
see Toni's remarks at 
https://github.com/scala/community-builds/pull/799#issuecomment-429091848 about the technical feasibility and required effort level of this

SethTisue:
some previous discussion on this at https://github.com/lightbend/dbuild/issues/193

SethTisue:
it seems unlikely this will progress, unless new impetus comes from outside. I don't think the ticket needs to stay open

SethTisue:
we ended up removing even Scala.js itself (see #1009), instead accomplishing equivalent testing here: https://github.com/scala-js/scala-js-test-scala-nightly

ticket on the possibility of taking that approach further: https://github.com/scala-js/scala-js-test-scala-nightly/issues/1

PR #799, opened 2018-10-10
[WIP] add scalajs-react

SethTisue:
references #798.  /cc @sjrd @japgolly


SethTisue:
queued: 
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3626/consoleFull

SethTisue:
in local test, the scalajs-react "core" project builds, but then when it tries to build "extra",

```
[scalajs-react] [warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[scalajs-react] [warn] 	::          UNRESOLVED DEPENDENCIES         ::
[scalajs-react] [warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[scalajs-react] [warn] 	:: com.github.japgolly.scalajs-react#core;1.3.2-dbuildx0ea4fff6fa5190a90a767f4c9af744e43fc61d0e: not found
[scalajs-react] [warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

I saw the same pattern in scala-js-dom, where building "root" works, but trying to build "extra" fails to find the dependency on the thing "root" published

@cunei it's tantalizing, it seems like this might be thiiiiiiiis close to working. any clue what's going on here?

cunei:
Hi @SethTisue, I investigated the matter a while back. If I remember correctly, the issue is in the generation of version numbers by Scala.js, which appends to the version some additional string in order to distinguish the JavaScript artifacts from the Java ones; I seem to remember that classifiers are involved, but I might be mistaken. While changing version numbers, dbuild does not know that the suffix is supposed to be part of the artifact name, and builds the artifact id in a manner that is not the same that Scala.js expects. Adding the needed support for Scala.js in dbuild is feasible, but requires handling Scala.js artifact names in an ad-hoc manner, meaning that some more code is needed.
Adding Scala.js support has not been a priority for the past few years; if you think that adding support would be important, you should ask for some development time to be devoted to that. A fair estimate would one to two weeks.

japgolly:
Hey @SethTisue, thanks for setting this up. LGTM

configs/community.dbuild
```diff
@@ -411,1940 +222,28 @@ build += {
...
-    name: "circe"
-    uri:  ${vars.uris.circe-uri}
-    extra.sbt-version: ${vars.sbt-1-version}
-    extra.projects: [
-      // easy
-      "coreJVM", "numbersJVM"
-      // harder
-      "jawn"
-      // bunch more stuff that all depends on jawn
-      "parserJ
```
japgolly, line 246: Can I ask, what's the reason for excluding these? Is it to save build time?

SethTisue: this is an include list, not an exclude list. `extra.projects` is include, `extra.exclude` is exclude.

typically we include what downstream projects actually depend on, and then the rest is gravy (we include or not based on some estimate of cost/benefit. most often the cost has to do with maintenance effort, not build time)

SethTisue: (btw I'm always happy, if someone asks, to add a comment to the config file clarifying a decision of this sort. you'll see there are already many such comments, but also many places where they are missing.)

japgolly: Ok cool thanks for clarifying. Can I confirm also that `test:compile` is run on the included projects? If it is then exclusion of the `gh-pages` module here is fine but if not; well preferably we'd switch it on for scalajs-react but if that's unfeasible for whatever reason then enabling `gh-pages` module would be the second-most valuable thing.

SethTisue: > Can I confirm also that `test:compile` is run on the included projects?

not only that, but `test:test`, too

japgolly: Awesome! Cheers

smarter:
> if you think that adding support would be important, you should ask for some development time to be devoted to that. A fair estimate would one to two weeks.

Is there anything that non-Lightbend people can do to encourage/help prioritize that ? Do +1 on https://github.com/scala/community-builds/issues/798 count ? I believe that this would be well worth the effort.

SethTisue:
> Is there anything that non-Lightbend people can do to encourage/help prioritize that ? Do +1 on #798 count ? I believe that this would be well worth the effort.

the one thing I can definitely say, that I've said in public a couple times before, is that any current or potential Lightbend customers with a commercial interest in Scala.js should definitely communicate that interest to the business and/or sales sides of Lightbend.

whether there are _other_ avenues by which this could be encouraged and helped, I really just don't know, it's above my pay grade, so to speak. regardless, I'd personally love to see this happen and I remain optimistic about it (as, admittedly, I have been for a long time now).

closing for now, just because I don't like to leave pull requests open indefinitely.

@cunei would it be worth asking @sjrd if he sees some way to work around this? I'm not clear enough on the artifact naming issue to know whether the suggestion even makes sense.

it would be valuable even if the workaround only worked for projects immediately downstream from Scala.js, we could worry later about whether it supported, for example, including projects which (for example) depended on scalajs-react

PR #802, opened 2018-10-16
move Akka forward to v2.5.17 and update test exclusions for JDK 11

SethTisue:
the goal here is to get Akka green in the JDK 11 build, since it's blocking a bunch of downstream projects

but we have to be sure it doesn't regress on 8

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -455,12 +455,11 @@ build += {
     ]
   }
 
-  // frozen (September 2018) at v2.5.16.  this repo is volatile (very
+  // frozen (October 2018) at v2.5.17.  this repo is volatile (very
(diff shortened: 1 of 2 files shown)


SethTisue:
* JDK 8: 
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3647/ (404 til queue clears)
* JDK 11: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-jdk11-integrate-community-build/95/ (404 til queue clears)

SethTisue:
verified in local run this gets akka itself (just the actors module, that is, not akka-more) green on JDK 11, but we still need to wait for the JDK 8 Jenkins run to see if anything breaks downstream

Issue #824, opened 2018-12-07
2.13 dropping symbol literals


ashawley:
I have a branch of Scala 2.13 that drops symbol literals:

https://github.com/scala/scala/pull/7499

There's a proposal to deprecate symbol literals in 2.13, but not to drop them.  So this is just an exploratory branch with no intent to merge in 2.13.  

In preparation for dropping after 2.13, would it be possible to see what this branch breaks in the community build? I predict since ScalaTest uses symbols and is a root dependency in the ecosystem, that not much of the question would be answered, but I figure it's worth a shot of letting the community build explode once and then inspecting the wreckage for clues.

Also, the branch introduces a new symbol shorthand syntax (`sym"foo"`), see https://github.com/scala/scala/pull/7495, which no projects will be using, but it would will be good to have community build double-duty check if it introduces problems.

SethTisue:
(I'll respond after my December vacation)

ashawley:
Happy vacationing!

ashawley:
Closing in favor of #827.

PR #827, opened 2018-12-28
2.12 Use alternate symbol syntax

ashawley:
Previously, in #824:

> There's a proposal to deprecate symbol literals in 2.13, but not to drop them. So this is just an exploratory branch with no intent to merge in 2.13.
>
> In preparation for dropping after 2.13, would it be possible to see what this branch breaks in the community build? I predict since ScalaTest uses symbols and is a root dependency in the ecosystem, that not much of the question would be answered, but I figure it's worth a shot of letting the community build explode once and then inspecting the wreckage for clues.
>
> Also, the branch introduces a new symbol shorthand syntax (sym"foo"), see scala/scala#7495, which no projects will be using, but it would will be good to have community build double-duty check if it introduces problems.

I toyed around with the community build on my own and found that 2.13 didn't offer a lot of feedback.  It's still a work in progress until 2.13.0 is released.  It seems 2.12 might be a better place to experiment.

For 2.12, I've tried dropping symbol literals from the compiler in https://github.com/scala/scala/pull/7585.  Because this change touches a lot of the ecosystem, I discovered I had to change the dbuild configs and fork a number of community libraries.

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -731,7 +731,6 @@ build += {
   ${vars.base} {
     name: "sourcecode"
     uri:  ${vars.uris.sourcecode-uri}
-    extra.sbt-version: ${vars.sbt-0-13-version}
     // no Scala.js plz
     extra.projects: ["sourcecodeJVM"]
   }
@@ -751,14 +750,6 @@ build += {
     name: "fastparse"
     uri:  ${vars.uris.fastparse-uri}
     extra.sbt-version: ${vars.sbt-0-13-version}
-    extra.projects: [
-      "fastparseJVM"   // no Scala.js plz
-      "scalaparseJVM"  // dependency of scalatex
-    ]
-    extra.commands: ${vars.default-commands} [
-      // too prone to unexplained failure
-      "set executeTests in scalaparseJVM in Test := Tests.Output(TestResult.Passed, Map(), Iterable())"
-    ]
   }
 
   ${vars.base} {
@@ -1040,7 +1031,6 @@ build += {
   ${vars.base} {
     name: "scalamock"
     uri:  ${vars.uris.scalamock-uri}
-    extra.sbt-version: ${vars.sbt-0-13-version}
     extra.commands: ${vars.default-commands} [
       // work around https://github.com/scala/scala-dev/issues/252 ;
       // see https://github.com/scala/community-builds/issues/434
@@ -1048,9 +1038,8 @@ build += {
     ]
     extra.projects: [
       // no Scala.js
-      "scalamock-coreJVM"
-      // dependency of github4s
-      "scalamock-scalatest-supportJVM"
+      "scalamockJVM"
+      "examplesJVM"
     ]
   }
 
--- a/configs/project-refs.conf
+++ b/configs/project-refs.conf
@@ -11,8 +11,10 @@ vars.uris: {
   akka-http-session-uri:        "https://github.com/softwaremill/akka-http-session.git"
   akka-http-uri:                "https://github.com/akka/akka-http.git"
   akka-persistence-cassandra-uri: "https://github.com/scalacommunitybuild/akka-persistence-cassandra.git#community-build-2.12"  # was akka, 2a53c6a0a1f64
+  akka-more-uri:                "https://github.com/ashawley/akka.git#drop-symlit" # was akka, master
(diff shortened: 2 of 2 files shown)


adriaanm:
Wow! Thanks for the investigation. What's your sense of the magnitude of this change? How did you adapt your forks to the new syntax?

ashawley:
@adriaanm  In the forks, I've been replacing symbol literals with the `sym"id"` short-hand syntax from https://github.com/scala/scala/pull/7495.  I had to backport it to 2.12 in https://github.com/scala/scala/pull/7585 so I can test in the 2.12 community build.  

I've had a few passes here with running dbuild locally, and it already affects close to 20 projects.  Hoping to run it on Jenkins to confirm that and see if there are any more that surface.

I'd say that there aren't a lot of community projects using symbol literals, but the ones that are affected are Scala tools like test frameworks, so projects that in turn use those test frameworks are similarly impacted.  This has the butterfly effect where deprecating and dropping symbol literals needs to ripple through the community projects.  This is probably a common scenario for these types of changes to Scala that cascade like this.

SethTisue:
I merged origin/2.12.x onto your branch and triggered https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3937/ (404 til Jenkins rouses himself from his slumber)

ashawley:
Thanks, Seth.  How do I configure it to use the compiler artifacts from https://github.com/scala/scala/pull/7585?  I had been accomplishing it locally with `./run.sh -v 2.12.8-bin-87ed2be-SNAPSHOT`, but I couldn't figure out what Jenkins is expecting.

SethTisue:
oh sorry, I should have been paying closer attention. triggered a new job with the Scala version specified:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3942/ (404 til Jenkins wakes up)

SethTisue:
oh, ha, `2.12.9-bin-`, not `2.12.8-bin-`. https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3945/

SethTisue:
```
SUCCESS 92 FAILED 13 DID NOT RUN 85 TOTAL 190
UNEXPECTED 13
BLOCKERS:
62 shapeless
32 spray-json
19 akka
12 scalapb
3 coursier
2 grizzled
1 scalatex
1 lift-json
```

ashawley:
Thanks, Seth.  Could I have one more run?  I tried forking akka, spray-json and lift-json.

SethTisue:
I'd suggest that you merge 2.12.x onto your branch before we do the next run.

ashawley:
Ok, done.  

A lot of the things that you had to fix on the 2.12.x branch I already had pinned to forks I froze last month.  It also added some knew projects, so we'll see what happens.

ashawley:
Seth,

A lot of my forks are frozen from early December, so there's a good chance that that will cause problems.  I didn't think that it would be a problem for the 2.12 community build, but I'm seeing otherwise and you probably know very well it to be the case.   Bit rot is real.

I don't think I'll have the time to do another full iteration of this.  I was able to get pretty far.  You probably have more pressing 2.13 priorities.  With respect to learning more about dropping symbols, it seems like diminishing returns, anyway.

SethTisue:
okay, makes sense.

a piece of advice I should have thought to offer earlier: this repo has a `freeze` script at the root level that lets you freeze all the project SHAs at whatever some Jenkins run that you pick used. And it's also possible to just rewind to the last time we froze for a release and work from that. Either way, it stops everything from constantly changing out from under you.

SethTisue:
@ashawley did this experiment produce any conclusions that you should record at https://github.com/scala/scala/pull/7395 ? (which I'm still inclined to go forward with)

SethTisue:
> This has the butterfly effect where deprecating and dropping symbol literals needs to ripple through the community projects

fwiw, I'm comfortable with that. deprecate-in-2.13, drop-in-2.14 is a multi-year timetable. and by deprecating first, we don't break the community build and we don't prevent projects from publishing.

ashawley:
Yes, that would be the ideal timetable.  It's difficult though. When it is dropped in 2.14, those libraries that expose symbol literals will need to handle cross-compiling for both situations if they, and most likely they will, need to cross-publish for 2.12, 2.13 and 2.14.

PR #830, opened 2019-01-16
add testz

SethTisue:
fyi @edmundnoble @dwijnand


SethTisue:
already green in local testing

PR #838, opened 2019-01-17
unfreeze cachecontrol

SethTisue:
I suspect this will need play-ws changes (maybe just refresh our fork). let's find out

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -958,10 +958,9 @@ build += {
     extra.projects: ["play-jsonJVM"]  // no Scala.js plz
   }
 
-  // frozen (October 2018) at v2.0.0-M6 tag because that's what playframework
-  // currently uses. playframework and play-ws tend to get a bit out of sync
-  // with each other so it's generally better to use a tag for play-ws.
-  // temporarily forked (October 2018) to remove uses of DebugConfiguration which was
+  // playframework and play-ws tend to get a bit out of sync with each other so it's generally better
+  // to have play-ws frozen at tag, unless it's forked. which it is right now:
+  // temporarily forked (February 2019) to remove uses of DebugConfiguration which was
   // removed from ssl-config; https://github.com/lightbend/ssl-config/issues/142
   // is about restoring it
   ${vars.base} {
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -984,8 +984,6 @@ build += {
     extra.exclude: ["integration-tests"]
   }
 
-  // dependency of play-ws. frozen (June 2018) at a March 2018 commit
-  // before an incompatible change was made (removing joda-time dependency)
   ${vars.base} {
     name: "cachecontrol"
     uri:  ${vars.uris.cachecontrol-uri}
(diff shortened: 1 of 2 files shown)


SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3972/

SethTisue:
```
[play-ws] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.16/project-builds/play-ws-0810349f3fb97171cc26898b42a0ec3a75ee2356/play-ahc-ws-standalone/src/main/scala/play/api/libs/ws/ahc/cache/AhcHttpCache.scala:10:12: object joda is not a member of package org
[play-ws] [error] import org.joda.time.{ DateTime, Seconds }
[play-ws] [error]            ^
[play-ws] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.16/project-builds/play-ws-0810349f3fb97171cc26898b42a0ec3a75ee2356/play-ahc-ws-standalone/src/main/scala/play/api/libs/ws/ahc/cache/EffectiveURIKey.scala:11:12: object joda is not a member of package org
[play-ws] [error] import org.joda.time.DateTime
[play-ws] [error]            ^
```

SethTisue:
play-ws eliminated joda time in May 2018

I made a test branch `test-20190117` that combines play-ws master with the changes from my community-build-2.12 branch, let's see how it does

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/3985/

SethTisue:
@marcospereira @dwijnand this still isn't working, play-ws depends on an old version of cachecontrol, it isn't source-compatible. are you interested in doing the upgrade, or should I just shrug and stick with the old cachecontrol?

I imagine you might need to deal with this anyway as part of the general move onto Scala 2.13

dwijnand:
I don't know anything about this.

I'm happy if you pin this as needed.  If you could create upstream tickets we'll try and resolve them in due time.

SethTisue:
I'm happy to just leave cachecontrol frozen for now. presumably this will crop up again later as part of some future attempt to upgrade Play stuff here, we can deal with it then.

SethTisue:
I think we need to tackle this in order to get playframework green in the Scala 2.13 community build

SethTisue:
we need to resolve this in order to get playframework green in the 2.13 community build

it looks like the solution is not to track cachecontrol master, but to track 1.1.x (or freeze at cachecontrol 1.1.5), which has 2.13 support but doesn't have the joda change

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4085/ (404 til Jenkins queue clears)

SethTisue:
note that Play 2.7.0 uses play-ws 2.0.1, so that's a likely play-ws tag to try

dwijnand:
If you could detail either what fixes are needed and/or what the root problems are, I'll try and fix them.

Is this all from the mess up in ssl-config?

SethTisue:
> If you could detail either what fixes are needed and/or what the root problems are, I'll try and fix them

thanks, ball is in my court at the moment, but I'll let you know

> Is this all from the mess up in ssl-config

(not really, that's just one factor)

PR #835, opened 2019-01-16
add scala-pet-store

SethTisue:
fyi @pauljamescleary

PR #847, opened 2019-01-30
add doobie

SethTisue:
commit also includes some tweaks/fixes that happened to come up
while I was working on this


SethTisue:
spun off from #835, let's book the progress. /cc @tpolecat

PR #848, opened 2019-01-31
add tsec

SethTisue:
because scala-pet-store is going to need it


SethTisue:
split off from https://github.com/scala/community-builds/pull/835, already tested there

PR #851, opened 2019-02-01
unfork Akka for 2.13

SethTisue:
or rather, use the same fork we use in 2.12

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -488,8 +488,15 @@ build += {
     ]
   }
 
-  // 2.13: we're forked, and the fork is behind what we use in 2.12,
-  // but we should probably just leave it alone until the Akka team publishes for 2.13.0-M5
+  // frozen (January 2019) at v2.5.20.  this repo is volatile (very
+  // frequent commits) and when it needs to be rebuilt, the testing
+  // takes a long time (especially akka-more) and the downstream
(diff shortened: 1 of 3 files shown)


SethTisue:
but this will have to wait until I fix genjavadoc

SethTisue:
> but this will have to wait until I fix genjavadoc

I fixed genjavadoc today, but now scala-java8-compat is the blocker

SethTisue:
attempting to unfork for 2.12, first, at #865

SethTisue:
@ennru points out that Akka has already published for M5, so why wouldn't it be green here?

scala-java8-compat has published for M5 as well, though @Ichoran is still working on https://github.com/scala/scala-java8-compat/issues/113

it's now close enough to 2.13.0-RC1 release time that I should probably look for a way to work around this in the community build somehow, rather than wait for Rex (I already did such a workaround once, e.g. 
https://github.com/scala/scala-java8-compat/issues/113#issuecomment-414069372)

SethTisue:
We used a branch of Lukas's. The Akka team is working on publishing for RC1; once they publish we ought to be able to unfork.

SethTisue:
rather than keep this ticket open, I think it's fine to just assume I'll get to it in the course of my usual rounds

PR #853, opened 2019-02-04
Add network and disk requirements

ashawley:
This may be useful to disclaim to potential local runners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,9 @@ a corpus of Scala open source projects together using Lightbend's
 It's **3.2 million lines** of Scala code, total,
 from **185 projects** (as of January 2019),
 and takes about **15 hours** to run.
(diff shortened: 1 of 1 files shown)


ashawley:
@SethTisue These are my hunches.  Do you have better numbers?

SethTisue:
Thanks, but I'd like to keep the readme slender — just a few sentences to attract interest — and use the wiki for more in-depth info. Could you add this here instead? 👉 https://github.com/scala/community-builds/wiki/Local-runs

re: bandwidth, but I guess it's okay to say something like "several gigabytes", surely that's in the ballpark.

As for disk space, I'd say ~30G instead of ~20G

ashawley:
Yup, will add it to the Wiki. Thanks.

PR #882, opened 2019-03-28
add missing include filter in repo config

lrytz:


lrytz:
going for 2.12.x instead https://github.com/scala/community-builds/pull/883

PR #891, opened 2019-04-11
fix sconfig using jodersky spray-json fork

SethTisue:
fyi @ekrich

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -915,7 +915,9 @@ build += {
     uri:  ${vars.uris.spray-json-uri}
     extra.projects: ["sprayJsonJVM"]
     // otherwise sbt-gpg errors on `publish`
-    "set every gpgWarnOnFailure := true"
+    extra.commands: ${vars.default-commands} [
+      "set every gpgWarnOnFailure := true"
+    ]
   }
 
   // forked (September 2018) for JDK 11 friendliness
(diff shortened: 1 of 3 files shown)


SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4340/

ekrich:
Is the spray-json fork in the community build as well or does this just keep track of it as a `sconfig` dependency? 

Is the extra sbt setting something that could be changed in the build or is that a common thing?

SethTisue:
> Is the spray-json fork in the community build as well or does this just keep track of it as a `sconfig` dependency?

I'm currently trying to add the fork to the community build and use it on all projects that use spray-json, not just for sconfig. If that doesn't work out, I'll try using the fork in sconfig only.

> Is the extra sbt setting something that could be changed in the build or is that a common thing?

If you mean the change I made to the sconfig section, I took that out, I must have been confused and/or fat-fingered something, not sure what I was thinking.

SethTisue:
in a local test, `[info] Project sconfig---------------------: SUCCESS (project rebuilt ok)`, so then I just need to see if anything else breaks.

https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4342/

ekrich (commented):
I was referring to this change. I'm not sure how to do a review with the code highlighted. Anyway, this line. `"set every gpgWarnOnFailure := true"`

SethTisue:
oh, it's not important either way, feel free to set that in your build or not, it's fine. it's pretty common that we have to tweak this in the community build for repos that use sbt-gpg

SethTisue:
as an afterthought on this, it probably would have made more sense to set `organization` in the spray-json config section, rather than changing the organization in every downstream repo :-/

ekrich:
I guess I saw that but not all that familiar with the setup. There will be more opportunities for clean up I'm sure 😄

PR #900, opened 2019-05-03
(wip) sbt 1.3.0-M3

eed3si9n:


dwijnand:
You can bring back play-ws with https://github.com/dwijnand/play-ws/tree/avoid-sbt-1.3-regression or when https://github.com/playframework/play-ws/pull/338 is merged.

SethTisue:
is it too early to move the whole build to 1.3.0-RC1?

(if there are a handful of troublesome projects, we could override those to stick with 1.2.8 for now)

eed3si9n:
We should certainly try moving to 1.3.0-RC1, and fork the breakers if we have to.

SethTisue:
okay let's move to #909

PR #906, opened 2019-05-11
Update PureConfig config to use the official repo

ruippeixotog:
We removed Ammonite from our build quite some time ago, so there shouldn't be a problem with using the official repo directly now. Our builds on `master` rarely fail and when they do it's not because of compilation or unit test failures, so it should also be safe to use the `master` branch.

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -2277,14 +2277,9 @@ build += {
     uri:  ${vars.uris.circe-config-uri}
   }
 
-  # forked (November 2017, refreshed January 2018) to remove a build.sbt thing that
-  # adds Ammonite, because it isn't a true dependency and we don't want the complication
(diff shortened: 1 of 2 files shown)


SethTisue:
thank you!

SethTisue:
@ruippeixotog this is a problem for us:

```
  initialize := {
    val required = "1.8"
    val current = sys.props("java.specification.version")
    assert(current == required, s"Unsupported JDK: java.specification.version $current != $required")
  },
```

as we run the community build on JDK 11 and JDK 12 too, not just JDK 8

would it be possible to either remove the check, or make it possible to disable?

ruippeixotog:
Sure, this is old code that serves no purpose now. I just submitted https://github.com/pureconfig/pureconfig/pull/474.

SethTisue:
thank you!

Issue #897, opened 2019-04-28
projects are moving to Fastparse 2


SethTisue:
which is hard because it isn't remotely source-compatible with Fastparse 1

this is timely because right now is an important time to be testing on Scala 2.13, but if a project's 2.13 support is landing in a branch that also requires Fastparse 2, we have a problem

perhaps this can be addressed with splitting the build into spaces

or perhaps we could keep everything in the same space, but use dbuild to publish Fastparse 1 and 2 under different org and/or artifacts, and then override the dependency downstream accordingly

SethTisue:
this might be easier than I feared. eyeballing the config, it appears that fastparse is used by:

* coursier, scalatex, scala-xml-quote: these are leaf projects
* scalapb, scalameta (and scalapb may be a leaf other than being used in scalameta)

I can't tell just from eyeballing whether there are any projects that are downstream of scalapb and/or scalameta that might also use fastparse directly. but I do know that nobody uses scalapb who doesn't also use scalameta.

but in any case, anything downstream of scalameta is fairly leaf-ish, capstone stuff rather than fundamental stuff.

and anyway, I recently discovered over in the 2.13 build that scalameta is using a shaded version of fastparse 1, so we ought to be able to just add the repo with the shaded version to the build, and everything will be hunky dory! 🤞 probably any fastparse-1-using project will be able to use the shaded one too

SethTisue:
there's some question of whether I can track scalameta master, or whether I need to stick with the one I have. but I would probably be able to use the shaded version of fastparse 1 regardless

https://github.com/scala/community-builds/pull/910 was an attempt to move the 2.13 build to scalameta master; perhaps the difficulties I ran into there were 2.13-specific

/cc @olafurpg

SethTisue:
the fastparse repo still has an sbt build in it, but presumably the mill build is the main build these days?

SethTisue:
the fastparse repo doesn't have a proper sbt build anymore, there is a `build.sbt`, but it is vestigial and doesn't actually build the main code

SethTisue:
perhaps I should start by just fetching fastparse 2 from Maven Central, get everything working, and then worry about whether it's necessary to actually build it ourselves

SethTisue:
PR #915

olafurpg:
[...]

PR #910, opened 2019-05-21
2.13: try Olaf's scalameta branch

SethTisue:
cc @olafurpg

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1896,6 +1896,7 @@ build += {
     // use right version-specific source directories regardless of our weird dbuild Scala version numbers
     extra.commands: ${vars.default-commands} [
       "set unmanagedSourceDirectories in (commonJVM, Compile) += baseDirectory.value / \"scalameta\" / \"common\" / \"shared\" / \"src\" / \"main\" / \"scala-2.13.0-RC1\""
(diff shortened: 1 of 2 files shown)


SethTisue:
https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2175/ (404 til Jenkins wakes up)

olafurpg:
Note that the Scalameta build got simplified in https://github.com/scalameta/scalameta/pull/1863 by merging a lot of modules and also removing few unused modules.

SethTisue:
hello, rabbit? are you down here?

/me resumes digging

SethTisue:
scalameta uses its own shaded version of fastparse 1, which is okay, it's under a different org so it won't conflict with any other fastparse versions

but scalameta also uses scalapb, and the version of scalapb we have is too old (resulting in compile errors in scalameta), and the newer versions use fastparse 2

so in order to finish this, I think I will need to tackle #897 first (fastparse 2 upgrade)

SethTisue:
note also that I unforked scalapb but doing so lost the changes in the fork Adriaan made, to add the right Scala version specific source directories to `unmanagedSourceDirectories`, I'll need to cope with that in the usual way (it's better to do it in the dbuild config that handling it by forking)

SethTisue:
I think this will probably have to wait until June or July. @olafurpg thanks for doing the work that allowed us to get this far, at least

Issue #898, opened 2019-04-28
projects are moving to ScalaTest 3.1


SethTisue:
Scala 2.13 support is landing in important projects on branches that have upgraded to ScalaTest 3.1, e.g. in the Typelevel ecosystem

not sure whether addressing this might require making separate 3.0/3.1 spaces, or whether the change can be worked through without that, with a little help from maintainers. I'll see how it goes

the good news is that ScalaTest modularized out its ScalaCheck dependency, so once we get on 3.1, we'll be in a better place going forward

I have some wip at https://github.com/SethTisue/community-builds/commits/scalatest-3.1

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4432/

SethTisue:
note to self, note the recently merged PRs at https://github.com/scalatest/scalatestplus-scalacheck/pulls?q=is%3Apr+is%3Aclosed

SethTisue:
let's track this on #913 rather than keeping the issue open

PR #913, opened 2019-05-25
[WIP] use ScalaTest 3.1 branch

SethTisue:
previously #898 

the repos that are likely to break are those that use ScalaTest's ScalaCheck integration, which has moved to a separate repo (scalatestplus-scalacheck) and new package (`org.scalatestplus`)

moving to cats master (2.x) will probably be necessary

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -206,13 +206,26 @@ build += {
   }
 
   // see also scalatest-tests
-  // forked for: build tweak, JDK 11 friendliness, comment out a test
-  // fork refreshed (from 3.0.x branch) January 2019
   ${vars.base} {
     name: "scalatest"
     uri:  ${vars.uris.scalatest-uri}
-    extra.sbt-version: ${vars.sbt-0-13-version}
     extra.projects: ["scalatest", "scalactic"]
+  }
+
+  ${vars.base} {
+    name: "scalatestplus-junit"
+    uri:  ${vars.uris.scalatestplus-junit-uri}
+  }
+
+  ${vars.base} {
+    name: "scalatestplus-testng"
+    uri:  ${vars.uris.scalatestplus-testng-uri}
+  }
+
+  ${vars.base} {
+    name: "scalatestplus-scalacheck"
+    uri:  ${vars.uris.scalatestplus-scalacheck-uri}
+    extra.projects: ["scalatestPlusScalaCheckJVM"]
   }
 
   ${vars.base} {
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -227,6 +227,11 @@ build += {
     name: "scalatestplus-scalacheck"
     uri:  ${vars.uris.scalatestplus-scalacheck-uri}
     extra.projects: ["scalatestPlusScalaCheckJVM"]
+    // scala.MatchError: ff6da566+20190524-1028-SNAPSHOT (of class java.lang.String)
+    //   at com.typesafe.dbuild.project.build.LocalBuildRunner$.prepareDepsArtifacts(LocalBuildRunner.scala:106)
+    extra.settings: ${vars.base.extra.settings} [
+      "version in ThisBuild := \"1.0.0\""
+    ]
   }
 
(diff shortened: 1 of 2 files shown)


SethTisue:
a recent test run is https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4541/

SethTisue:
`SUCCESS 96 BLOCKERS: 52 catalysts, 21 play-json, 17 scodec-bits, 15 akka-stream, 13 scalapb, 10 jackson-module-scala, 9 jawn-0-10, 5 unfiltered, 3 slick, 2 scala-logging, 2 jsoniter-scala, 1 scalatestplus-scalacheck`

SethTisue:
an increasing number of projects are now frozen because they've moved to 3.1, so the pressure is gradually building on this. I'm reluctant to do it until 3.1 is out of SNAPs and into RCs or even final release

SethTisue:
maybe it wouldn't be so bad to leave the 2.12 build on 3.0 indefinitely, and only bite this bullet in the 2.13 context

another thought is that perhaps we could do this without spaces by overriding the artifact name for one of the two versions, and then override the dependency as needed downstream

(as an aside, I was also thinking it wouldn't be so bad if one of the two versions came from Maven Central rather than being built from source, but upon further reflection, I don't think there would be any big savings to doing that, we might as well build them both from source)

ashawley:
I was trying to troubleshoot ScalaCheck issues in the community build, https://github.com/rickynils/scalacheck/issues/488, but it seems the ScalaTest 3.1 is a cascading blocker for moving ScalaCheck, at all.  I'd suggest getting the community build to 3.1 by hook or by crook.

ashawley:
> I'm reluctant to do it until 3.1 is out of SNAPs and into RCs or even final release

This is my sense of the crux of the matter, as well.  Some projects just write their test suite in ScalaTest, so they wouldn't mind snapshot releases and could be convinced.  But there are projects that publish their own bindings for use in ScalaTest.  It would be hard to convince those projects to upgrade their code to ScalaTest 3.1 for the community build when it would have a transitive dependency to a snapshot release.

SethTisue:
>  I'd suggest getting the community build to 3.1 by hook or by crook.

yeah. it's the next big thing I'll tackle once #946 lands

SethTisue:
P.S. I'm increasingly thinking that I should just leave the 2.12.x and 2.11.x builds on ScalaTest 3.0, freezing projects as needed to keep it that way, and do the 3.1 upgrade in the 2.13.x context only.

SethTisue:
this looks timely: https://github.com/scalatest/scalatest/commit/eaa224b5b65f7a2b19962dd8e4bd188f4943956c ("Added 2.13.0 build to Travis")

SethTisue:
hmm but according to https://github.com/scalatest/scalatest/issues/1656 , there will be a 3.1.0-M1 soon, maybe I should wait for that

SethTisue:
I fooled with a 2.13 version for a bit today at https://github.com/SethTisue/community-builds/commits/scalatest-3.1-2.13-aug-2019

and older attempt is at https://github.com/SethTisue/community-builds/commits/scalatest-3.1-2.13

in the latest attempt,

```
[info] Project cloc-plugin-----------------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-----------------------: SUCCESS (unchanged, not rebuilt)
[info] Project scala-xml-------------------: SUCCESS (unchanged, not rebuilt)
[info] Project scalacheck------------------: SUCCESS (unchanged, not rebuilt)
[info] Project scalatest-------------------: SUCCESS (unchanged, not rebuilt)
[info] Project scalatestplus-junit---------: SUCCESS (project rebuilt ok)
[info] Project scalatestplus-scalacheck----: FAILED (RuntimeException: )
[info] Project scalatestplus-testng--------: SUCCESS (project rebuilt ok)
```

scalatestplus-scalacheck hasn't been updated in a few months, so it looks to me like I should just be patient and wait for 3.1.0-M1

PR #915, opened 2019-05-25
support fastparse 2 (instead of or in addition to fastparse 1), and unfreeze scalameta

### LINKED ISSUE #897: projects are moving to Fastparse 2

SethTisue:
attempt to fix #897

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -766,10 +766,11 @@ build += {
     extra.projects: ["utestJVM"]
   }
 
-  ${vars.base} {
+  // just fetch Fastparse 2 from Maven Central, since it has no sbt build anymore
+  {
     name: "fastparse"
-    uri:  ${vars.uris.fastparse-uri}
-    extra.exclude: ["readme"]  // out of scope
+    system: aether
+    uri: "aether:com.lihaoyi#fastparse_2.12;2.1.2"
   }
 
   // frozen (January 2019) at December 2018 commit before the build was changed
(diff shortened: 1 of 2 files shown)


SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4546/ (404 til Jenkins queue clears)

SethTisue:
```
[coursier] [error] org.eclipse.jgit.api.errors.JGitInternalException: Missing unknown 7b3678abc22e81e6cdb1ba804608178d0dbc1798
```

SethTisue:
one of the roadblocks here was that scalameta needed fastparse 1

but it's about time we unfroze scalameta, we're frozen at a version that is now quite old.

but actually scalameta depends on a shaded version of fastparse 1:

```
build.sbt:      // NOTE(olafur): use shaded version of fastparse because the latest version v2.0.0
build.sbt:      "org.scalameta" %%% "fastparse" % "1.0.1"
```

so perhaps that opens a door for us to support both 1 and 2 here! that would be ideal

SethTisue:
fixed by #946 in 2.13.x

in 2.12.x, I'm content to leave things alone, freezing projects as needed

PR #916, opened 2019-05-25
[2.12.x] Dedup alterKeyImpl

dwijnand:


dwijnand:
With some type backflips we could handle the duplicate calls to `in` (see `Scoped.ScopingSetting`), but handling `~=` requires handling the difference between transforming a bare `T` value and transforming a value within a `Task[T]` (and preserving the resulting shape) and for that there isn't any directly supporting code.  So just instanceOf/asInstanceOf pattern match our way out of finding that solution.

dwijnand:
WDYT @SethTisue?

SethTisue:
I’ll get to it :-)

dwijnand:
image

SethTisue:
DON'T YOU DARE DELETE THE BRANCH I KNOW YOU WANT TO

dwijnand:
Let's come back to this one.

SethTisue:
resubmission: #940

PR #929, opened 2019-06-21
Add a compile-time only dependency of jsoniter-scala

plokhotnyuk:


SethTisue:
optimistically merging. if there turns out to be a problem we can circle back

SethTisue:
will the 2.12.x branch need this too?

plokhotnyuk:
yes, and 2.11.x too

SethTisue:
okay, I applied the change to all three branches.

(the way we currently do this is to make changes on 2.12.x first, then merge them onto 2.11.x and 2.13.x)

plokhotnyuk:
@SethTisue Thank you a lot! I have missed that the 2.12.x is the default one...

plokhotnyuk:
Also, the missing part was checking for JDK 8 during publishing... I have fixed [here](https://github.com/plokhotnyuk/expression-evaluator/commit/6a469abf8a4f523aa05fe544dfa72aef396b0c23) to allow publishing with JDK 11

PR #934, opened 2019-07-09
remove obsolete test excludes

SethTisue:
well, we'll see what's obsolete and what isn't...


SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4664/

SethTisue:
all were still green except `scribe`

SethTisue:
this was just a single test run on JDK 8. so, we'll see if some of these exclusions need to come back, but commented as "intermittent" or "JDK 11 only" this time

PR #940, opened 2019-07-24
Dedup alterKeyImpl

SethTisue:
resubmission of #916, because @dwijnand put a pillow over its face
while it was sleeping


SethTisue:
optimistically merging, as one does

PR #941, opened 2019-07-24
keep the Scala SHA in a properties file

SethTisue:


dwijnand:
TIL properties files (or a subset of property files?) are sourceable!

SethTisue:
only a subset, yeah. (in particular, if the property name has a `.` in it, which typical Java properties are likely to, it doesn't work. and of course there's a whole universe of other things that won't work.)

Issue #936, opened 2019-07-10
move to scalameta master (or v4.2.0 tag) and enable `testsJVM`


SethTisue:
see https://github.com/scala/community-builds/issues/935 for background

> move to scalameta 4.2.0 with `testsJVM` enabled (making or requesting any needed fixes), scalafmt master, and then if scalafix breaks as a result, just remove or disable it until @olafurpg (or whoever) has moved it to scalameta 4.2.0 as well

SethTisue:
this is currently blocked because scalameta master uses a version of scalapb that depends on fastparse 2

scalameta still uses fastparse 1 internally, but it uses a shaded version that we can retrieve with aether, so we won't have a fastparse 1 vs 2 conflict, but adding fastparse 2 is still work. (it's #897)

olafurpg:
See scalameta/fastparse for the manually shaded version of fastparse v1. It might be possible to add it to the community build since the package name doesn’t conflict with fastparse v2.

SethTisue:
wip on this at #946

SethTisue:
#946 fixes this on 2.13.x and will be merged soon

in 2.12-land, I think I'm content to leave things where they are.

PR #946, opened 2019-07-31
2.13: support both fastparse 1 and 2; get scalameta green again; unfork ScalaPB; drop moribund projects

### LINKED ISSUE #897: projects are moving to Fastparse 2

SethTisue:
sequel to #910

fixes #897, #915, #936

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -2075,29 +2075,6 @@ build += {
     check-missing: false  // ignore missing scalafmt
   }
 
-  // dependency of autowire. (and Ammonite, but we pulled Ammonite.)
-  // otherwise obsolete, as per https://github.com/lihaoyi/upickle-pprint/issues/209
-  // (the pprint part has its own repo now)
-  // requires jawn 0.10
-  // April 2018 news: upickle has been revived, and is now
-  // a dependency of Haoyi's new uJson library, which we
-  // should probably add at some point. for now, doing nothing
-  ${vars.base} {
-    name: "upickle"
-    uri:  ${vars.uris.upickle-uri}
-    extra.sbt-version: ${vars.sbt-0-13-version}
-    // no Scala.js; also only upickle no pprint
-    extra.projects: ["upickleJVM"]
-  }
-
-  // dependency of scaladex. depends on upickle
-  ${vars.base} {
-    name: "autowire"
-    uri:  ${vars.uris.autowire-uri}
-    extra.sbt-version: ${vars.sbt-0-13-version}
-    extra.projects: ["autowireJVM"]  // no Scala.js plz
-  }
-
 ]}
 
 //// space: jawn_0_11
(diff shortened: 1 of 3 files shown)


SethTisue:
queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2407/

SethTisue:
with Maven-retrieved (rather than source-built) fastparse 2, in local testing:

```
[info] Project scalameta-------------------: SUCCESS (project rebuilt ok)
```

but the question is what else will break.

SethTisue:
queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2411/

SethTisue:
I wrote new sbt builds for fastparse 2 and utest (since they only have mill builds these days). Doing so was reasonably easy/pleasant.

And in local testing now, all the low-level projects I would want to still be green, are. So I'm now confident this should go forward; if there ends up being more breakage farther downstream, we'll deal with it.

I'm a bit tempted to try to target 2.12.x on this. But I think there's no real need; in general, it's fine for the 2.12.x to get increasingly frozen. If it eventually turns out that other 2.12.x community build things would end up benefiting from this, we can do the backport then.

SethTisue:
Added a commit which drops nearly all of the projects in the "jawn 0.10" space, including sbt and its modules. jawn 0.10 is unlikely to ever make the jump to 2.13. /cc @eed3si9n

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2415/

eed3si9n:
I created a new release of sjson-new recently using Jawn 0.14.2 for Scala 2.13 - https://github.com/eed3si9n/sjson-new/blob/60158cbc3a42ef235b8c85390da8ef6ec6f28a4a/project/Dependencies.scala#L25

SethTisue:
@eed3si9n I'll attempt to re-add

overall PR situation: latest test run has `SUCCESS 131 FAILED?! akka-http,upickle,scalatex` which is excellent. I think the akka-http failure is unrelated, and getting upickle and scalatex green again probably won't be hard (and in any case isn't completely essential). so this is nearly ready for merge

SethTisue:
* scalatex: fixed by forking it
* upickle: dropped
* sjson-new: re-added
* akka-http: fixed (scalacommunitybuild/akka-http@215e6744d490354ea64638123c72cc0ef4a41c50)

PR #947, opened 2019-08-01
Improve format of report output

ashawley:
I have feedback about the output of the report as a newcomer.  The numbers for blockers are strangely placed, the use of the noun "SUCCESS" is imprecise, the longer lines are hard to read, and the choice of punctuation (?!) was confusing.

After toying around with it, I made some changes. 

If the report format has a basis in Jenkins or any other purposes, then feel free to decline.

### DIFF
--- a/report/Report.scala
+++ b/report/Report.scala
@@ -79,17 +79,16 @@ object SuccessReport {
             blockerCounts(blocker) = 1 + blockerCounts.getOrElse(blocker, 0)
       }
     val total = success + failed + didNotRun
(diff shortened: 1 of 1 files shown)


ashawley:
Here's an abridged version of the output as an example:
```
SUCCEEDED: 18
FAILED: play-json,cats
BLOCKERS: play-json (4), cats (4), twitter-util (3)
FAILED: 3
DID NOT RUN: 6
TOTAL: 27
```

SethTisue:
thanks!

PR #954, opened 2019-08-09
newer http4s SHA

SethTisue:
goals here:
1) get on same SHA as 2.13 build has, which is good because
1a) reduce unnecessary 2.12/2.13 differences
1b) it's newer, better to move forward if we can, even though we can't unfreeze entirely until we have ScalaTest 3.1
2) see if EntityDecoderSpec still fails, as it has in recent runs
   (probably ScalaCheck's fault, but we can pursue that next)

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -2366,9 +2366,9 @@ build += {
 
   // this is a big build with a ton of subprojects.  we needn't be shy
   // about disabling fragile ones, ones with obscure dependencies we don't have, etc.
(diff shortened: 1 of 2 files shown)


SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4816/

SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4818/

SethTisue:
`EntityDecoderSpec` still fails. will deal with that separately

PR #950, opened 2019-08-03
[WIP] add lagom samples

SethTisue:
sequel to #869

PR #962, opened 2019-08-16
unfreeze lagom, add akka-management

SethTisue:
leading up to #950, hopefully

cc @dwijnand 

build:
* https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4892/


SethTisue:
@patriknw in the akka-management repo, are there any tests I ought to actually run and not just compile?

patriknw:
> in the akka-management repo, are there any tests I ought to actually run and not just compile?

@SethTisue I don't think there are much interesting unit tests to run there. The more interesting tests are integration tests requiring more infrastructure and moving parts than I think should be part of community build

PR #965, opened 2019-08-20
unfreeze twitter-util + finagle

SethTisue:
let's see if the same `@varargs` failure (a possible scalac bug) we're seeing in 2.13 appears here as well. if it doesn't show up on JDK 8, try JDK 11 too


SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4917/

SethTisue:
@martijnhoekstra your 2.13 twitter-util PR undid https://github.com/twitter/util/commit/7e5024bd42327e6acae7954ac9b212b5ed51610d; curious why it came up?

martijnhoekstra:
I shouldn't have undone that -- I was trying to make some `@varargs` overloads work, and I slapped in on more things than I should have. Did it break something?

After some discussion I came to the conclusion that `@varargs` on a parameter list before the last one is a bug, and leads to useless bytecode, but I didn't see breakage.

I have a branch lying around that disallows the annotation in scala, but I figured the change is so minimal that nobody is going to care anyway, and probably isn't worth the bandwidth of scala/scala

martijnhoekstra:
(Some) more details here: https://github.com/twitter/util/pull/248#discussion_r300328046

I have a branch lying around where I started to remove it, https://github.com/scala/scala/compare/2.13.x...martijnhoekstra:varargsLastOnly?expand=1 but I figured that having the annotation in that place is useless and mostly harmless, and breaking it for everyone who has that annotation on a useless place may be less desirable than living with the misfeature that does no major harm.

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4925/

> Did it break something?

yeah, but only on JDK 11+. I'll try to isolate it and file a scala/bug ticket

SethTisue:
green on JDK 8, so I'm going to merge, then deal with the JDK 11 thing separately

Issue #966, opened 2019-08-21
push sbt 1.3.0-RC4 upgrade through


SethTisue:
including re-testing all projects that failed on RC3

/cc @eed3si9n @eatkins

SethTisue:
this is now done on 2.12 as of https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4923/

the full list of projects that didn't work on RC4 is:

* fs2 (worked on RC3)
* genjavadoc
* scala-js
* blaze
* sbt-io, sbt-util, sbt-librarymanagement (very old versions)
* sttp
* tsec

Error details are in comments in https://github.com/scala/community-builds/blob/2.12.x/configs/community.dbuild

I'm working on pushing the change through the 2.11 and 2.13 builds.

eatkins:
I think sttp will pass with -Dsbt.classloader.close=false. We should probably bump the sbt pinned versions to something more recent. The tsec tests passed when I ran them locally so it may have just been a flaky test or something about the community build. The genjavadoc tests worked for me with 2.12.8 but failed in the same way the community build did with 2.13.0.

SethTisue:
> I think sttp will pass with -Dsbt.classloader.close=false

let's find out! https://github.com/scala/community-builds/pull/968 (UPDATE: green, merged)

> We should probably bump the sbt pinned versions to something more recent

I think the 1.3.0 release would be a good opportunity for that, but I'm hoping @eed3si9n will tackle it for me :-)

SethTisue:
on 2.13, scala-parallel-collections weirdness: f09999ddaccdf659f3758a0bf573d90daf9d8c50

PR #968, opened 2019-08-21
use sbt.classloader.close to avoid NoClassDefFoundError

SethTisue:
as suggested by @eatkins at
https://github.com/scala/community-builds/issues/966#issuecomment-523635824

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -1614,6 +1614,8 @@ build += {
     extra.options: [
       // hopefully avoid intermittent OutOfMemoryErrors with default heap
       "-Xmx2g"
+      // needed on sbt 1.3 to avoid NoClassDefFoundError after tests run
+      "-Dsbt.classloader.close=false"
     ]
     extra.commands: ${vars.default-commands} [
       // tests in these subprojects are too slow and (more importantly) too fragile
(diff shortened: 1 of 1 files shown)


SethTisue:
~~https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4928/~~
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/4929/

smarter:
... but what does that option do, and why would be it be necessary in these projects ?

SethTisue:
let's fyi @adamw (sttp) and @dwijnand (lagom) since they may hit this if they try to move their projects to sbt 1.3

eatkins:
@smarter sbt 1.3.0 closes the in process classloaders after test and run completes. This allows the classloaders to release their native file handles. Unfortunately, if the user code uses ShutdownHooks or leaks a thread, this can cause a crash after test or run completes. The documentation has explicitly warned users to not use ShutdownHooks with in process test and run, but a number of builds do it anyway. In the case of sttp, they don't directly use ShutdownHooks, but I suspect that they don't properly shutdown the actor systems that they start up in test which causes both a thread leak and a shutdown hook leak.

SethTisue:
@eatkins that'd be a super addition to https://www.lightbend.com/blog/sbt-1.3.0-release

SethTisue:
green in local testing

dwijnand:
What's the downside of upgrading to sbt 1.3 and using `-Dsbt.classloader.close=false`?

eatkins:
The benefits of _not_ setting `-Dsbt.classloader.close=false` are:
1. we release the file handles to jars. Keeping the handles open can prevent these jars from being deleted on windows. This probably doesn't matter on mac and linux unless the system is starved for file descriptors.
2. It can help identify thread leaks which ultimately lead to sbt running out of memory fairly quickly

I've gone back and forth on what the right default is.

dwijnand:
Thanks.

PR #590, opened 2017-08-14
add squants

SethTisue:
keeping up with the Dotty community build Joneses

PR #974, opened 2019-09-10
add squants

SethTisue:
previous attempt #590, maybe it'll go better this time

fyi @cquiroz


SethTisue:
gah

```
[squants:error] <set>:5: error: value extract is not a member of object Project
[squants:error]     val extracted = Project extract s
[squants:error]                             ^
[squants:error] <set>:7: error: value relation is not a member of object Project
[squants:error]     val r = Project.relation(extracted.structure, true)
[squants:error]                     ^
[squants:error] <set>:15: error: not found: value structure
[squants:error]     BuiltinCommands.reapply(session, structure, s)
[squants:error]                                      ^
```

not clear why only this repo would be hitting this, what's special about it? I'll have to dig around

SethTisue:
oh, I think it's because `project/build.scala` has `object Project { ...`

SethTisue:
Yup, that was it. Green in local testing.

PR #975, opened 2019-09-13
exclude various intermittent failures

SethTisue:
affected projects:
* kafka
* scalatest-tests
* cats
* akka / akka-stream
* scalikejdbc
* finagle
* http4s

### DIFF
--- a/configs/community.dbuild
+++ b/configs/community.dbuild
@@ -385,6 +385,10 @@ build += {
     // just fluctuate :-/  I've been gradually increasing this trying to fix an OOM in
     // ApplicativeSuite.  the repo's own .jvmopts has 6g
     extra.options: ["-Xmx6g"]
+    extra.commands: ${vars.default-commands} [
+      // intermittent failures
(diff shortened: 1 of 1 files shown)


SethTisue:
fyi @ennru note that I stopped running the tests in entire subprojects, excluding individual tests wasn't working, even after endless cycles, intermittent failures remained

fyi @kailuowang on the cats exclusion, I've seen this fail many times

fyi @rossabaker on the http4s exclusion

fyi @seratch on scalikejdbc

fyi @bvenners on scalatest-tests (I'm on the verge of just giving up and only compiling your tests and not running them; perhaps you could put all of the asynchronous tests in a different subproject, since those are the ones that seem to fail?)

fyi @patriknw note that I stopping running any `remote` tests at all, the failures were too many/various

SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5028/

bvenners:
@SethTisue We'll try and tag the flickering ones and then you can exclude those when you run.

SethTisue:
@bvenners another possibility would be to wait until ScalaTest 3.1 ships, then revisit. right now the community build is still on 3.0.x

PR #980, opened 2019-09-23
merge 2.12.x to 2.13.x, unfreeze, refreeze

SethTisue:
notably, this adds the `configs/unfrozen-refs.conf` file from 2.12.x (but with appropriate 2.13.x contents), so we have both the frozen and unfrozen states always in version control


SethTisue:
queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2631/

PR doesn't include a refreeze commit yet (since the freeze script needs a Jenkins run to use as reference)

SethTisue:
I'm hitting a tough problem with scala-parallel-collections where if I use sbt 1.3.2, junit somehow isn't on the classpath, but I can't downgrade to 1.2.8 because the latest mima doesn't work there

it were some random third-party project I'd probably just freeze or fork it, but scala-parallel-collections is one where it seems pretty important that we track master

I will dig into the junit classpath thing and try and find a solution or workaround, I already tried and failed once a week or two ago, but I'll try again.

SethTisue:
I found a workaround, `useCoursier := false`, surely I tried this at least once before? but perhaps with some other combination of settings that made it fail? anyway, seems to be okay for now, 🤞

SethTisue:
notes to self

# needs revisiting/reporting

* scala-parallel-collections freeze

# done

## akka

test failure reproducible outside dbuild, reported upstream
worked around it by excluding test

# maybe done

# scalameta

weird coursier dependency problem, let's just freeze scalameta
and worry about it some other time

## monix

https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2641/artifact/logs/monix-build.log

[monix] [error] **** Missing dependency: the library io.monix#implicitbox is not provided (in space "scala.main") by any project in this configuration file.
[monix] [error] In order to control which version is used, please add the corresponding project to the build file


ah, just need to add this:

  // dependency of monix
  ${vars.base} {
    name: "implicitbox"
    uri:  ${vars.uris.implicitbox-uri}
    extra.projects: ["implicitboxJVM"]
  }


## twitter-util

weird `ParRange` compilation error

investigation steps:

* try to reproduce outside of dbuild
  * first try 2.13.1: `++2.13.1` `util-stats/Test/compile`
    * result: compiles
  * then try 2.13 nightly: `++2.13.2-bin-520a88e`
    * result: compiles
  * then switch to sbt 1.3.2
  * try 2.13.1 again: `++2.13.1` `util-stats/Test/compile`
    * result: compiles
  * then try 2.13 nightly: `++2.13.2-bin-520a88e`
    * result: compiles
* okay, let's see what we can do within dbuild
  * first cut the config file down:
    * depends on: cloc-plugin, jackson-module-scala, scala, scala-collection-compat, scala-parallel-collections, scala-parser-combinators, scala-xml, scalacheck, scalatest
    * just build util-stats, not other subprojects
    * run it and verify it fails. DONE
  * now try rewinding both t-u and s-p-c to sbt 1.2.8

possible next steps:
* rewind twitter-util to old sbt?
* rewind s-p-c to old SHA + sbt 1.2?
  * if we use old sbt, we have to either rewind SHA or fork to get rid of mima
  * rewinding all the way to v0.2.0 runs into a source incompatibility
  * let's try rewinding to effa33489bc02723a1aba1e36911128a500dfc68 (May 2019)
    which is where the needed source change was made
  * aha, that fixed it!  okay, so continue bisecting
    * use `extra.test-tasks: ["compile"]` to speed up bisect
    * using `git bisect`, but manually pasting the SHAs into configs/project-refs.conf,
      I don't think it's worth automating?
    * the bisect is a bit tricky because I need to adjust the s-p-c entry in the config
      file to include the right workarounds for the sbt version in question
  * gah, during bisect I ran into `java.lang.NoClassDefFoundError: xerial/sbt/Sonatype$`
    during dependency extraction of s-p-c. under what circumstances does this happen?
    I still don't know, despite having seen it a lot.  I'll try `git clean -ffxd` --
    but does that fix it? no. neither does setting sbt version to 1.3.0. I guess it's
    a problem with `sbt-scala-module" % "2.1.0"`?  I'll `git bisect skip` it

```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
156ed3073bda636b51cbeeca7072acc9a63d14b0
dab76f9c69bcf0a45ae8904acc8469ae4d8bdaee
a2dd4c1ced7d89af34fd8781ea8e8d6a190285e2
We cannot bisect more!
```

what about looking at the s-p-c JAR and seeing if the classes are missing?

hmm, what about trying to reproduce it outside dbuild using `publishLocal`?

# scala-js

https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2641/artifact/logs/scala-js-build.log

```
[scala-js] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.16/project-builds/scala-js-a976b60efd6a19472ddcddc97bf7c03e98baaa8f/linker/jvm/src/main/scala/org/scalajs/linker/frontend/optimizer/ParIncOptimizer.scala:40:44: Symbol 'type scala.collection.generic.ParMapFactory' is missing from the classpath.
[scala-js] [error] This symbol is required by 'object scala.collection.parallel.mutable.ParTrieMap'.
[scala-js] [error] Make sure that type ParMapFactory is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[scala-js] [error] A full rebuild may help if 'ParTrieMap.class' was compiled against an incompatible version of scala.collection.generic.
[scala-js] [error]     def emptyParMap[K, V]: ParMap[K, V] =  ParTrieMap.empty
[scala-js] [error]                                            ^
```

huh, this looks similar to the twitter-util thing!

# fast-string-inter​polator

https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2641/artifact/logs/fast-string-interpolator-build.log

```
[fast-string-interpolator:error] <set>:1: error: not found: value bintrayReleaseOnPublish
[fast-string-interpolator:error] bintrayReleaseOnPublish := false
```

# avro4s

https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2641/artifact/logs/avro4s-build.log

wtf is this

```
[avro4s] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.16/project-builds/avro4s-6ffddf2be2c194c8810f3d8e50b63abc26c2ad0b/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/EnumDecoderTest.scala:17:47: class AvroEnumDefault does not extend scala.annotation.Annotation
[avro4s] [error] case class ScalaAnnotatedSealedTraitEnumList(@AvroEnumDefault(List(CuppersAnnotatedEnum)) cupcat: List[CupcatAnnotatedEnum])
[avro4s] [error]                                               ^
```

can we reproduce outside dbuild? `avro4s-core/test:compile`

* with 2.13.1: yes, reproduced

upstream PR submitted

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2644/

SethTisue:
rebased.

split the scala-parallel-collections trouble out as #981

SethTisue:
geting close! `FAILED: akka-http,monix `

https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2644/artifact/logs/monix-build.log monix needs older cats, I think?

https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2644/artifact/logs/akka-http-build.log akka-http test failure, perhaps intermittent? but seen twice so far, maybe it's consistent

SethTisue:
queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2648/

PR #1004, opened 2019-10-17
advance SHAs for Akka and related repos

SethTisue:
Akka 2.5.26 was just released, so it seems like a good time do
this


SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5198/

PR #1005, opened 2019-10-18
wip for scala-js#3805

SethTisue:
https://github.com/scala-js/scala-js/pull/3805


SethTisue:
queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/2730/

SethTisue:
looks like overriding the Scala version works partially but not completely, note that `Non-compiled module 'compiler-bridge_2.12' for Scala 2.13.2-bin-99afc54`

```
[scala-js] [info] Compiling 1 Scala source to /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.16/project-builds/scala-js-5cf745dc8badcad5d60888296f563ceed652299d/junit-async/jvm/target/scala-2.12/classes ...
[scala-js] [info] Non-compiled module 'compiler-bridge_2.12' for Scala 2.13.2-bin-99afc54. Compiling...
[scala-js:error] /tmp/sbt_36080b2b/xsbt/Compat.scala:18: error: object IR is not a member of package scala.tools.nsc.interpreter
[scala-js:error]   val Results = scala.tools.nsc.interpreter.IR
[scala-js:error]                                             ^
[scala-js:error] /tmp/sbt_36080b2b/xsbt/ConsoleInterface.scala:11: error: object ILoop is not a member of package scala.tools.nsc.interpreter
[scala-js:error] import scala.tools.nsc.interpreter.{ ILoop, IMain, InteractiveReader, NamedParam }
[scala-js:error]        ^
```

@sjrd @gzm0 maybe we can figure out some other approach.

might we be able (once you've merged the PR and done a release) to retrieve binary artifacts for the subprojects in question from Maven Central and use those, instead of building them?

it might also work to split scala-js.conf in two. I can try that next (but not this week)

sjrd:
> looks like overriding the Scala version works partially but not completely

Perhaps we'd need to set it in the `Compile` configuration of each project, in addition to just the projects?

> might we be able (once you've merged the PR and done a release) to retrieve binary artifacts for the subprojects in question from Maven Central and use those, instead of building them?

No, that would not work, because every time we change the IR and hence the compiler and linker, we absolutely need the new linker defined in the new commit. The old linker on Maven Central wouldn't be able to read the IR produced by the new compiler.

> it might also work to split scala-js.conf in two. I can try that next (but not this week)

That's an interesting approach.

SethTisue:
there is some further discussion at https://github.com/scala-js/scala-js/pull/3805

but the whole thing may be moot: https://github.com/scala-js/scala-js/pull/1009

PR #1006, opened 2019-10-23
fix scala-parser-combinators on JDK 14

SethTisue:
because I love debugging random OSGi lossage 😍


SethTisue:
"the SethTisue:jdk14-2.11 branch can be safely deleted." — and cast into Mount Doom, preferably

PR #1018, opened 2019-12-13
sbt 1.3.5 (was 1.3.4)

SethTisue:
fyi @eed3si9n

### DIFF
--- a/advance
+++ b/advance
@@ -1,4 +1,4 @@
-#!/usr/bin/env sbt -Dsbt.version=1.3.4 -Dsbt.main.class=sbt.ScriptMain -Dsbt.supershell=false -error
+#!/usr/bin/env sbt -Dsbt.version=1.3.5 -Dsbt.main.class=sbt.ScriptMain -Dsbt.supershell=false -error
 
 // to advance all projects:
 //   % ./advance scalacheck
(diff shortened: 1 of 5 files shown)


SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5320/

SethTisue:
```
[twirl] [info] Twirl compiler
[twirl] [info] - should compile successfully (real) *** FAILED ***
[twirl] [info]   play.twirl.compiler.test.Helper$CompilationError: Symbol 'type scala.xml.NodeSeq' is missing from the classpath.
[twirl] [info] This symbol is required by 'value play.twirl.api.BaseScalaTemplate.x'.
[twirl] [info] Make sure that type NodeSeq is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[twirl] [info] A full rebuild may help if 'BaseScalaTemplate.class' was compiled against an incompatible version of scala.xml.
[twirl] [info] (NOTE: It looks like the scala-xml module is missing; try adding a dependency on "org.scala-lang.modules" : "scala-xml".
[twirl] [info]        See http://docs.scala-lang.org/overviews/ for more information.)
```

SethTisue:
@eed3si9n I dealt with the Twirl failure by using 1.3.4 instead in that one repo. does it seem like something you might want to investigate...?

eed3si9n:
Yea. That sounds like a regression.

PR #1033, opened 2019-12-30
2.13: support both ScalaTest 3.0 and 3.1

SethTisue:
this will probably take weeks of intermittent work to work through...

fixes #1020 

note to self: update dependencies.txt before merging

PR #1040, opened 2020-01-14
2.13: sweeping upgrade, build-wide

### LINKED ISSUE #1020: 2.13: upgrade to ScalaTest 3.1
there is previous history on this at #913, #898, #875 

as discussed there, I think it's fine for the 2.11 and 2.12 builds to stay on ScalaTest 3.0 forever, and we can just freeze things downstream as needed

but the 2.13 build, we should start moving now that 3.1 is in final release. we have already accumulated a substantial number of frozen projects (some green but frozen, some known-to-fail) that are waiting on this. (I've commented them with "ScalaTest 3.1" so I can find them)

and also as discussed at #913 already, I think the best strategy is:
* build 3.1 from source
* for projects still on 3.0, use 3.0 binaries from Maven Central. it's fine

SethTisue:
summary:

* ScalaTest 3.1 is now supported (fixes #1020)
* Major projects unfrozen
* Formerly failing repos now pass
* New repos added
* Some repos dropped
* Some notable repos still fail
* No more "spaces"

These are likely the last major changes the Scala 2 community build will see.

(the wip on this PR was at #1033)

## ScalaTest 3.1 is now supported

We build both 3.0 and 3.1 from source.  3.1 is built normally.  3.0 is
built from a fork where the artifacts' organization name is changed to
`scalacommunitybuild`. That allows both versions to coexist in the
same dbuild "space".

3.1 is the default.  At present, we override the dependencies of
26 repos to point to 3.0 instead.

Lack of support for ScalaTest 3.1 had long blocked many other
improvements, as detailed below.

## Major projects unfrozen

We had frozen many repos, for many months in some cases, because they
required ScalaTest 3.1 support, and then often downstream repos had to
be frozen too, as a result.  All of these repos are now using current
code again if reasonably possible.

Some especially notable upgrades include:

* cats: now at 2.1.x
* cats-effect: now at 2.0
* fs2: now tracking master
* scalameta: now 4.3.0

## Formerly failing repos now pass

* algebra
* circe-jackson
* discipline-scalatest
* doobie
* kittens
* meow-mtl
* metrics-scala
* multibot
* scalafmt
* tsec

## New repos added

There was no goal here to expand the overall size of the build, but
the following additions happened anyway, because they were needed as
dependencies:

* cats-retry
* cats-testkit-scalatest
* cats-time
* circe-generic-extras
* ciris
* context-applied
* discipline-specs2
* http4s-jwt-auth
* jwt-scala
* log4cats
* natchez
* redis4cats
* scalatestplus-junit
* scalatestplus-scalacheck
* scalatestplus-testng
* scodec-cats
* skunk
* sttp-model

## Some repos dropped

Keeping these was problematic for one reason or another, but we don't
need them:

* catalysts
* circe-derivation
* scrooge-shapes

And the following repos never made the jump to 2.13 at all. We can
always re-add the if they someday do:

* curryhoward
* paradox
* scalastyle

## Some notable repos still fail

With luck, we will be able to get these green soon:

* coursier
* eff (upstream needs cats 2.1 upgrade)
* pfps-shopping-cart
* scalafix

## No more "spaces"

For a long time, the 2.13 build has been closely aligned with the
2.12 build to facilitate forward merges.  This PR marks the end
of that close alignment in a few significant ways.

First, as described above, we no longer include all the same repos as the2
2.12 build; repos that never made the jump to 2.13 have been removed.

Second, we no longer make use of dbuild's "spaces" feature. The 2.12 build
uses it to keep jawn 0.10 and jawn 0.11 separate, but the 2.13 build
doesn't contain any 0.10-based projects.

This simplifies maintenance, since `community.conf` is now a
simple alphabetically-ordered list.


SethTisue:
this will probably need at least some light followup work on JDK 11 and JDK 13, but I didn't want to hold the PR up over that

SethTisue:
fyi @bvenners @cheeseng

fyi @tpolecat @gvolpe @TravisBrown @OlegYch @olafurpg @ChristopherDavenport @vlovgr @augustjune @profunktor @pauldijou @djspiewak

proj/scalastyle.conf
```diff
@@ -1,9 +0,0 @@
-// https://github.com/scalastyle/scalastyle.git#master
```
mwz, line 1: Hi Seth, Scalastyle hasn't been maintained for a long time (https://github.com/scalastyle/scalastyle/issues/327) and the maintainers are no longer active, so I decided to [fork](https://github.com/beautiful-scala/scalastyle) the project and [release](https://github.com/beautiful-scala/scalastyle/releases/tag/v1.1.0) a new version with support for 2.13. I also updated and published the [website](https://scalastyle.beautiful-scala.com).

Would it be possible to add this project to the community build instead?

SethTisue: Sure, let's give it a try: https://github.com/scala/community-build/pull/1055

PR #1047, opened 2020-01-14
2.12: advance project SHAs

SethTisue:


SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5395/

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5397/

SethTisue:
* circe: compilation failure in jawn subproject. either freeze circe, or freeze jawn?
* akka-http: freeze before ScalaTest 3.1 upgrade
* jawn-fs2: compile errors. additional evidence jawn needs freezing?
* scalike-jdbc: freeze before ScalaTest 3.1 upgrade
* scala-debugger: test failure, perhaps intermittent?
* case-app: ambiguous implicits in shapeless-using code, not sure why, maybe just freeze at known green?

SethTisue:
we already have case-app forked/frozen, so let's try freezing shapeless instead

~~https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5399/~~ https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5402/

SethTisue:
@alexarchambault @milessabin the shapeless / case-app thing might interest you — I dealt with it by freezing shapeless. error log is at https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/5397/artifact/logs/case-app-build.log

milessabin:
@SethTisue It's been a while, and I've lost track of how this works. Are you tracking shapeless master or are you on the current shapeless release?

If it's the former then it's most likely fallout from work towards the next, bincompat breaking, release. /cc @joroKr21

alexarchambault:
I guess it's https://github.com/milessabin/shapeless/pull/797 that now conflicts with [the](https://github.com/alexarchambault/case-app/blob/4d1b6414a13975a9b1e354fef1b5585eaca4f013/core/shared/src/main/scala/caseapp/core/parser/HListParserBuilder.scala#L64) [workarounds](https://github.com/alexarchambault/case-app/blob/4d1b6414a13975a9b1e354fef1b5585eaca4f013/core/shared/src/main/scala/caseapp/core/parser/LowPriorityHListParserBuilder.scala#L12) that case-app had.

These workarounds will likely go away when case-app switches to the upcoming shapeless version, which should fix this error.

joroKr21:
Yeah that sounds about right

SethTisue:
>  It's been a while, and I've lost track of how this works. Are you tracking shapeless master or are you on the current shapeless release?

re: "how this works", it's totally on a repo-by-repo basis (and may also be different in our 2.11, 2.12, and 2.13 branches)

in the Shapeless repo, we normally track master, which has the advantages that we get a collective heads-up if you do something that breaks something downstream, like now, or if you do something that isn't compatible with a recent Scala nightly

it would also be perfectly fine for me to pin Shapeless at a release version, which has the advantage of reducing effort/noise for both of us.

so, advantages either way, either way is reasonable.

2.12 Shapeless config currently is https://github.com/scala/community-build/blob/2.12.x/proj/shapeless.conf

milessabin:
@SethTisue I think the problem with pinning on a non-release, non-master commit is that you'd be testing against a neither here nor there state. I'm not sure what value it would offer.

I'm tracking shapeless master against scalac nightlies [here](https://github.com/milessabin/shapeless-test-scala-nightly), so I'm reasonably confident that I'll be notified promptly if a scalac change breaks things.

shapeless 2.4.0 will potentially break downstream projects, but that's expected.

So on balance I think it would be better to pin to 2.3.3.

@joroKr21 what do you think?

milessabin:
Alternatively, are there any workarounds for @alexarchambault's workarounds that would allow us to keep going with master?

SethTisue:
> So on balance I think it would be better to pin to 2.3.3

On the `shapeless-2.3.3` tag, `crossScalaVersions` has `2.13.0-M2` which is so ancient that I suspect using that tag wouldn't work.

Would it be appropriate to use the `shapeless-2.3.3-scala-2.13.0-RC3` branch, or is there something better?

(As an aside: this sort of confusion is why I'm typically fine with having a repo in a "neither here nor there state" as long as everything is green. Regardless, happy to use whatever branch, tag, or commit you suggest.)

joroKr21:
I don't know if there is any workaround of the workaround. I don't expect that much breakage for 2.4 (afaiu there are no bincompat problems with dbuild) but I expect 2.5 to break a lot.

milessabin:
> On the shapeless-2.3.3 tag, crossScalaVersions has 2.13.0-M2 which is so ancient that I suspect using that tag wouldn't work.

Ahh ... right. Hmm ...

joroKr21:
Btw if we merge milessabin/shapeless#807 there will be no need for a workaround at all.

The only downside is that tagged types will start boxing.
But considering how many problems they create this might be acceptable.

PR #1058, opened 2020-01-22
unfork refined

SethTisue:


SethTisue:
queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/3090/

SethTisue:
oh, right, I forgot, we can't do this because the version of discipline-scalatest needed is too new. (moving to that newer version might work, but there's not really a reason I need to pursue it at the moment)

PR #1059, opened 2020-01-23
2.12: new Scala SHA

SethTisue:


SethTisue:
queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5421/

SethTisue:
hmm, 5421 and 5423 both have sttp and scala-refactoring failing

it seems implausible that the sttp test failure could be related, but it has happened twice in a row (and it isn't the usual shutdown errors that repo intermittently shows)

perhaps the scala-refactoring thing is an overly sensitive test, but perhaps not

SethTisue:
I was able to reproduce the sttp failure outside of dbuild on 2.12.10. no clue why we only started it seeing it here, now, but anyway, I'll just exclude the test. here on 2.12.x we're using a pretty old sttp commit so it doesn't seem worth reporting upstream, either

SethTisue:
as for scala-refactoring: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/5427/artifact/logs/scala-refactoring-build.log

the `inlineInListConcatenation` and `testConsOperator` failures both seem genuine (as in, rewrites that once produced correct code are now producing incorrect code), and they are reproducible outside dbuild with `++2.12.11-bin-0efdbf7`

note that are no maintainers to ask to help, the repo is abandoned.

the test failures have to do with rewriting of `::` and `:::`, so it seems nearly certain that the relevant PR is https://github.com/scala/scala/pull/8618 . I'll ask Harrison about it there

SethTisue:
in order not to hold up the PR, I excluded the failing scala-refactoring tests

PR #1061, opened 2020-01-24
Remove expression-evaluator from the `jdk14Failures` list

plokhotnyuk:
Now it compiles even with JDK 15.


SethTisue:
cool, I'll take your word for it and we can revisit if Jenkins turns out to disagree

PR #1107, opened 2020-03-11
2.12: new Scala SHA

SethTisue:
this is the SHA we expect to release as 2.12.11


SethTisue:
optimistically merging because we only merged one very safe-seeming change

PR #1125, opened 2020-04-06
add scala-rewrites

SethTisue:
fyi @dwijnand

queued https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5657/

### DIFF
--- a/proj/scala-rewrites.conf
+++ b/proj/scala-rewrites.conf
@@ -27,8 +27,10 @@ vars.proj.scala-rewrites: ${vars.base} {
   extra.commands: ${vars.base.extra.commands} [
     "removeDependency org.scalameta semanticdb-scalac"
     "removeDependency ch.epfl.scala scalafix-testkit"
+    "removeDependency ch.epfl.scala scalafix-cli"
     """set input / libraryDependencies += compilerPlugin("org.scalameta" % "semanticdb-scalac_2.12.11" % "4.3.7")"""
     """set tests / libraryDependencies += compilerPlugin("org.scalameta" % "semanticdb-scalac_2.12.11" % "4.3.7")"""
     """set tests / libraryDependencies += "ch.epfl.scala" % "scalafix-testkit_2.12.11" % "0.9.13" % Test"""
+    """set tests / libraryDependencies += "ch.epfl.scala" % "scalafix-cli_2.12.11" % "0.9.13" % Test"""
   ]
 }
--- a/proj/scala-rewrites.conf
+++ b/proj/scala-rewrites.conf
@@ -5,12 +5,20 @@
 // can't upgrade it because it has too may dependencies.  (And
 // scala-rewrites is 2.12-only because it lives in sbt-plugin-land.)
 
+// This is trickier-than-normal because semanticdb-scalac and
+// scalafix-testkit are CrossVersion.full, so we have to override that
+// to use older binaries, since binaries don't exist for Scala nightlies.
+//
+// (If those older binaries don't work, then we'll have to do
+// something else, but in a way that's good because we can notify Olaf
+// upstream that we changed something in Scala that's going to require
+// changes to Scalafix.)
+
 vars.proj.scala-rewrites: ${vars.base} {
   name: "scala-rewrites"
   uri: "https://github.com/scala/scala-rewrites.git#82a063a7fefe91a13d0c3bf754c4a057cfb4c579"
 
-  // extra.exclude: ["output213"] // ???
-  extra.exclude: ["tests", "input"]
+  extra.exclude: ["tests"] // TODO
   // the use-binary-dependencies thing runs into trouble because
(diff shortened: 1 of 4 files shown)


SethTisue:
@dwijnand looking at the errors, my first guess is that the scalafix version in the 2.12 community build is way, way too old. and scalafix sits on top of such a tall tower of dependencies that I don't think we can upgrade it

normally at this point I would give up, but scala-rewrites is of fundamental importance. however, it's not clear to me whether having it in the community build matters or not. we could still add it, in its own space, using dependencies pulled from Maven Central. is it clear to you whether that's worth doing...?

note that this alternate approach would be equivalent to adding "latest Scala nightly" to Scala-rewrites' own CI — wdyt about that? (sometimes people feel weird about having their CI build become non-reproducible that way)

dwijnand:
Given scala-rewrites isn't released yet, I'm not pressed for it to be in the community build.  I don't feel weird about adding a "latest Scala nightly" variant to its CI, if you want to PR that.

SethTisue:
>  I don't feel weird about adding a "latest Scala nightly" variant to its CI, if you want to PR that.

I tried, but It's a bit tricky because `semanticdb-scalac` and `scalafix-testkit` are `CrossVersion.full`, so you get e.g. "ResolveException: Error downloading org.scalameta:semanticdb-scalac_2.12.12-bin-085b6cf:4.3.7". We could get around it by rigging something that uses the 2.12.11 versions instead on the assumption that they'll probably work...

...but that, plus the fact that https://github.com/sbt/sbt/pull/5464 hasn't landed (and I never plugin-ized [this advice](https://github.com/scala/scala-dev/issues/648#issuecomment-525923361) which means it has to be pasted in every repo)...

...has me wanting to try the add-it-to-the-community-build-using-binary-dependencies-route next instead. (I'll hit the same `CrossVersion.full` problem there, but I feel more comfortable putting a grungy workaround in `proj/scala-rewrites.conf` than I do putting it in the scala-rewrites repo itself.)

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5665/

proj/scala-rewrites.conf
```diff
@@ -0,0 +1,36 @@
+// https://github.com/scala/scala-rewrites.git#master
+
+// This project is unusual in that we build it using only binary
+// dependencies, since the community build's scalafix version is too
+// old and we can't upgrade it because it has too many dependencies.
+// (And scala-rewrites is 2.12-only because it lives in
+// sbt-plugin-land.)
```
dwijnand, line 7: sbt-plugin-land?  It's 2.12 only because Scalafix is 2.12 only, for some reason:

https://repo1.maven.org/maven2/ch/epfl/scala/scalafix_2.13/

proj/scala-rewrites.conf
```diff
@@ -0,0 +1,36 @@
...
+
+// This is trickier-than-normal because semanticdb-scalac and
+// scalafix-testkit are CrossVersion.full, so we have to override that
+// to use older binaries, since binaries don't exist for Scala nightlies.
```
dwijnand, line 11: Perhaps we can get dbuild / the community build to version the scala artefacts as `-bin-...`?

proj/scala-rewrites.conf
SethTisue: I thought Scalafix was 2.12 only primarily because it typically lives in sbt-land, but perhaps I'm mistaken.

proj/scala-rewrites.conf
SethTisue: not sure what you mean

dwijnand: Ah, wait (_he says to himself_) it's `CrossVersion.patch` that would work better here, right?

proj/scala-rewrites.conf
dwijnand: You might be better informed than me.  I thought it was 2.12 for the longest time because Scalameta was 2.12 only, and now it's 2.12 only because no one's got round to upgrading it.

proj/scala-rewrites.conf
SethTisue: still not sure what you mean

dwijnand: OK, sorry.  What I'm trying to say is how do we avoid having to do the `removeDependency` and `set input / libraryDependencies += ...` in the community build?  I think the answer is `CrossVersion.patch`, but I reckoned you would know best.

SethTisue: `CrossVersion.patch` strips off the `-bin-` suffix, but we'd still be left with e.g. 2.12.12, which doesn't exist yet

dwijnand: Ah, right.  Damn.

PR #1144, opened 2020-05-03
2.13: new Scala SHA

SethTisue:
~~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1440/~~
~~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1441/~~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1456/


SethTisue:
the change to `Symbol#toString` means some projects need tweaking (either in the code, or just by excluding tests). working through this is tedious but it's fine

one of the projects needing tweaks was Shapeless, and the repo already has the needed changes, so I unfroze to get them, but I realized today as I got deeper into this that the post-2.3.3 Shapeless changes also remove support for symbol literals, which then hits a number of downstream projects, and the whole thing starts getting complicated (or if not actually complicated, at least, time-consuming)

so I think I now need to back up and fork Shapeless from our existing freeze point

SethTisue:
* Shapeless: resolved
* bug I found in ScalaTest: fixed (but I should remember to submit the fix upstream)

of the remaining failures, there are a couple that I think just involve updating expected outputs (or excluding tests): scalatest-tests, treehugger

but there is a real problem with pureconfig (causing both pureconfig-more and coulomb to fail). outside dbuild, in the pureconfig repo with `++2.13.3-bin-43b70d4` and `tests/console`,

```
scala> import pureconfig._, pureconfig.generic.auto._; case class Conf(x: Int)

scala> ConfigWriter[Conf].to(Conf(3))
val res1: com.typesafe.config.ConfigValue = SimpleConfigObject({"ymbol-(x-)":3})
```

note the `ymbol`. I'm still digging, but I'm beginning to suspect the problem may be in Magnolia rather than in Pureconfig itself. or possibly even further down the stack in Shapeless itself

SethTisue:
ah, it's in pureconfig

```diff
-      val keyStr = hint.configKey(key.value.toString().tail)
+      val keyStr = hint.configKey(key.value.name)
```

SethTisue:
disabling running scalatest-tests tests for now, too many are essentially testing `Symbol#toString` directly

PR #1148, opened 2020-05-13
unfreeze scala-rewrites

SethTisue:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/5752/

### DIFF
--- a/proj/scala-rewrites.conf
+++ b/proj/scala-rewrites.conf
@@ -1,7 +1,4 @@
-// https://github.com/scala/scala-rewrites.git#82a063a7fefe91a13d0c3bf754c4a057cfb4c579  # was master
-
-// frozen (April 2020) at a known-green commit before fix.RuleSuite
-// started failing; see https://github.com/scala/scala-rewrites/issues/19
+// https://github.com/scala/scala-rewrites.git#master
 
 // This project is unusual in that we build it using only binary
 // dependencies, since the community build's scalafix version is too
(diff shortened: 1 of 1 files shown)


SethTisue:
@dwijnand what do you make of these test failures? https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/5752/artifact/logs/scala-rewrites-build.log

SethTisue:
I did a local run with `./narrow scala-rewrites` and `version=2.12.11` and got the same failures. so, doesn't have to do with post-2.12.11 changes, apparently

dwijnand:
> what do you make of these test failures?

My first guess is the "Power" thingy isn't kicking in? https://github.com/scala/scala-rewrites/pull/14/files#diff-b09e11820eea2f4cca90f9346899e64a

> I did a local run with `./narrow scala-rewrites` and `version=2.12.11` and got the same failures.

I'm stating the obvious, but 2.12.11 runs cleanly on Travis CI: https://travis-ci.com/github/scala/scala-rewrites/builds/162320176 so it seems to do with either the modded config in the CB, or something deeper in the CB?

Thanks for that check though.  I was going to suggest that I try to setup Travis CI cron jobs to run against 2.12.next (next CB't candidate release version).  Now it seems not relevant here, so I'll postpone that until it proves worthwhile.

About this one, I'll have to circle back to go deeper to understand what's going on.  Thanks for the initial push.

dwijnand:
I've found a way to test Scala nightlies in sbt proper, in scala-rewrites:

* https://github.com/scala/scala-rewrites/pull/35

and I've set up nightly builds of it:

https://travis-ci.com/github/scala/scala-rewrites/settings

So we can drop this effort, IMO.

PR #1165, opened 2020-06-21
new Scala 2.12 SHA

SethTisue:


SethTisue:
this is green except for scala-collection-laws. merging anyway, I'll look into the failure in time for 2.13.3

PR #1167, opened 2020-06-22
new Scala 2.13 SHA

SethTisue:
and undo a few now-unneeded forkings


SethTisue:
I'm going to go ahead and merge this since it's getting close to release time and we don't have enough disk space to keep so many different Scala versions in the dbuild caches

PR #1168, opened 2020-06-22
new Scala 2.12 SHA (& reinstate linter + multibot)

SethTisue:


SethTisue:
merging, having verified locally that linter is green, since it's close to 2.13.3 release time and we're low on disk space

PR #1186, opened 2020-07-22
2.13: advance project SHAs (and remove scala-gopher)

SethTisue:
~~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1650/~~
~~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1651/~~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1654/


SethTisue:
@rssh I'm removing scala-gopher from the 2.13 build, since it requires the old scala-async. (frozen versions of scala-gopher will remain in the 2.11 and 2.12 builds.)

it wouldn't be impossible for the community build to support both old and new scala-async, but doing that would require substantial additional effort, so I prefer not to do that.

if you ever move scala-gopher to the new scala-async, let me know and we can try re-adding it to the 2.13 build.

SethTisue:
the other failures in run 1651 are apparently because of a version mismatch between pureconfig upstream, and coulomb and refined-more downstream

UPDATE: this was easily resolved by freezing pureconfig at 0.13.0 for now

rssh:
Thanks, an update of scala-gopher to new async is in plans, but hard to say when.  I will submit the patch to re-add when will do that.

PR #1190, opened 2020-07-26
reinstate scalatest-tests

SethTisue:


SethTisue:
not mergeable yet — waiting on scalatest/scalatest#1851, scalatest/scalatest#1852, scalatest/scalatest#1853, scalatest/scalatest#1854

SethTisue:
rather to my surprise, this was green in local testing

https://github.com/scalatest/scalatest/pull/1854 has the 2.13.3 upgrade, but we currently don't run the tests, just compile them, so that isn't a hold-up. we can re-enable running the tests later

I don't know why we don't hit https://github.com/scalatest/scalatest/pull/1852 here — presumably something order-dependent in the build

PR #1191, opened 2020-07-30
run scalatest-tests again

SethTisue:
now that https://github.com/scalatest/scalatest/pull/1855 is merged, this might work? let's find out


SethTisue:
green in local test

PR #1210, opened 2020-09-02
Remove publish from droste.conf excludes

eatkins:
Project extraction failed with 1.4.0-SNAPSHOT.

### DIFF
--- a/proj/droste.conf
+++ b/proj/droste.conf
@@ -12,7 +12,7 @@ vars.proj.droste: ${vars.base} {
 
   extra.exclude: [
     // out of scope
-    "*JS", "docs", "readme", "publish", "coverage"
+    "*JS", "docs", "readme", "coverage"
     // [error] **** Missing dependency: the library io.github.stanch#reftree
     "reftree"
   ]


eatkins:
I'm not sure if this is a good idea or not but prior to this change, I got the following error trying to extract the project with 1.4.0-SNAPSHOT:
```
[droste] [error] These subprojects were not found: "publish".  Found: "root", "docs", "laws", "tests", "reftreeJS", "metaJS", "coverage", "athemaJS", "coreJS", "macrosJS", "lawsJS", "meta", "testsJS", "scalacheck", "macros", "athema", "readme", "core", "reftree", "scalach
eckJS".
```

SethTisue:
The version of droste we are building definitely has a `publish` subproject that we want to exclude. after `hub clone higherkindness/droste; cd droste; git checkout 1e4b3569fc30f8027f59e2c188bb739b5552271b; sbt projects` I see `publish` on the list. but if I change `project/build.properties` to `1.4.0-M2` and run `sbt projects` again, `publish` has vanished.

So this looks like a regression of some kind in 1.4.0-M2. (Perhaps there's some issue with the project having the same name as the `publish` task? Wild guess.)

eatkins:
Ok, thanks for the clarification. I do think it's a poor idea to have a project mirror a built-in task but I guess it should still work if it did before.

eatkins:
Actually, the problem was that droste was depending on undefined behavior in sbt: https://github.com/higherkindness/droste/pull/180.

PR #1213, opened 2020-09-09
2.12: add scala-collection-compat's `binary-compat` project

SethTisue:
waiting on https://github.com/scala/scala-collection-compat/pull/362

dependencies.txt will need an update since this adds a dependency on mima-core


SethTisue:
green in local test

PR #1235, opened 2020-10-01
re-enable Lagom tests

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1938/

### DIFF
--- a/proj/lagom.conf
+++ b/proj/lagom.conf
@@ -27,7 +27,10 @@ vars.proj.lagom: ${vars.base} {
     "set `persistence-cassandra-scaladsl` / Test / executeTests := Tests.Output(TestResult.Passed, Map(), Iterable())"
     "set `testkit-scaladsl` / Test / executeTests := Tests.Output(TestResult.Passed, Map(), Iterable())"
     // more flaky tests I haven't reported upstream
-    "set `server-scaladsl` / Test / unmanagedSources / excludeFilter := HiddenFileFilter || \"LagomApplicationSpec.scala\""
+    """set `server-scaladsl` / Test / unmanagedSources / excludeFilter := HiddenFileFilter || "LagomApplicationSpec.scala""""
+    // these fail with mixed Akka versions, despite our multiple efforts above?!
+    """set `cluster-core` / Test / unmanagedSources / excludeFilter := HiddenFileFilter || "ClusterMessageSerializerSpec.scala""""
+    """set `persistence-jdbc-core` / Test / unmanagedSources / excludeFilter := HiddenFileFilter || "SlickOffsetStoreSpec.scala""""
   ]
   // as usual with scripted-plugin, sigh
   deps.ignore: ["org.scala-sbt#scripted-plugin"]
--- a/proj/lagom.conf
+++ b/proj/lagom.conf
@@ -20,15 +20,14 @@ vars.proj.lagom: ${vars.base} {
     // needed on sbt 1.3 (and 1.4? not sure) to avoid NoClassDefFoundError after tests run
     "-Dsbt.classloader.close=false"
   ]
-  extra.test-tasks: ["compile"]  // https://github.com/scala/community-build/issues/1227
   extra.commands: ${vars.default-commands} [
     // affects forked tests
     """set ThisBuild / javaOptions += "-Dakka.fail-mixed-versions=false""""
     // tests in these subprojects are too slow and (more importantly) too fragile
-    // "set executeTests in `persistence-cassandra-scaladsl` in Test := Tests.Output(TestResult.Passed, Map(), Iterable())"
-    // "set executeTests in `testkit-scaladsl` in Test := Tests.Output(TestResult.Passed, Map(), Iterable())"
+    "set `persistence-cassandra-scaladsl` / Test / executeTests := Tests.Output(TestResult.Passed, Map(), Iterable())"
+    "set `testkit-scaladsl` / Test / executeTests := Tests.Output(TestResult.Passed, Map(), Iterable())"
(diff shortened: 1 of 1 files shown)


SethTisue:
somehow our disabling of the akka-mixed-versions error isn't taking effect:

```
[lagom] [info] ClusterMessageSerializerSpec:
[lagom] [info] com.lightbend.lagom.internal.cluster.ClusterMessageSerializerSpec *** ABORTED *** (22 milliseconds)
[lagom] [info]   java.lang.IllegalStateException: You are using version 2.6.0 of Akka, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. You can solve this by adding an explicit dependency on version 2.6.0 of the [akka-discovery, akka-actor, akka-slf4j, akka-cluster-sharding-typed, akka-cluster, akka-distributed-data, akka-cluster-typed, akka-cluster-tools, akka-pki, akka-multi-node-testkit, akka-persistence, akka-cluster-sharding, akka-actor-typed, akka-coordination, akka-remote, akka-stream, akka-testkit, akka-serialization-jackson] artifacts to your project. See also: https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed
[lagom] [info]   at akka.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:217)
```

SethTisue:
@dwijnand can I ask you to take a look at this? (or, if you think it isn't worth the team's time, to say so? and/or to consult with your old team as needed?)

I [tried](b0b6920fc3aa8ad034b5048bee835a94c3b364ed) adding a couple specific test exclusions:

```
    """set `cluster-core` / Test / unmanagedSources / excludeFilter := HiddenFileFilter || "ClusterMessageSerializerSpec.scala""""
    """set `persistence-jdbc-core` / Test / unmanagedSources / excludeFilter := HiddenFileFilter || "SlickOffsetStoreSpec.scala""""
```

which worked but then when I hit yet another one with the same error:

```
[lagom] [error] Error during tests:
[lagom] [error] 	com.lightbend.lagom.internal.projection.ProjectionMessagesSerializerSpec
[lagom] [error] (projection-core / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

is when I started to feel like maybe we ought to either dig deeper into why the usual methods of suppressing these errors aren't working — or to throw up our hands, one of the two

dwijnand:
let's cc the team: @octonato / @ignasi35 any ideas why `set ThisBuild / javaOptions += "-Dakka.fail-mixed-versions=false"` isn't working as intended?

SethTisue:
Note that there are multiple other repos where the same `mixed-versions` incantations are needed and they successfully do the job: I had to do it akka-http-cors, akka-http-session, akka-http, akka, play-ws, playframework, sttp...

It's only lagom that is somehow different

ignasi35:
I don't know where it comes from but the following looks suspicious:

```
                    "project-specific-all-java-options-combined" : [
                        "-Xms1536m",
                        "-Xmx1536m",
                        "-Xss2m",
                        "-Doverride.akka.version=2.6.0"
                    ],
```

(copied from the [Jenkins output](https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1938/consoleFull)).

Why is `-Doverride.akka.version=2.6.0` necessary?

------------------

The code of `ManifestInfo` doesn't' change much. I checked the [history](https://github.com/akka/akka/commits/42223eb71adc5764a0fec91a95dee1702f6ef4ac/akka-actor/src/main/scala/akka/util/ManifestInfo.scala) of that file and could only find a couple of rather recent commits where the `IllegalStateException` is thrown from the line 217. It's only in [this commit] (https://github.com/akka/akka/blob/0e4d41ad33dbeb00b598cb75b4d29899371bdc8c/akka-actor/src/main/scala/akka/util/ManifestInfo.scala) and [this commit](https://github.com/akka/akka/tree/c41c0420ad1ed1841a47d5c4187e54b478936430/akka-actor/src/main/scala/akka/util/ManifestInfo.scala) that the exception is thrown from the line number 217, actually.

```
[lagom] [info] ClusterMessageSerializerSpec:
[lagom] [info] com.lightbend.lagom.internal.cluster.ClusterMessageSerializerSpec *** ABORTED *** (22 milliseconds)
[lagom] [info]   java.lang.IllegalStateException: You are using version 2.6.0 of Akka, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. ...
[lagom] [info]   at akka.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:217)
```

I assume this recent commits are in versions greater than `2.6.0` so I don't think `-Doverride.akka.version=2.6.0` is kicking in.

So, if `override.akka.version=2.6.0` is not kicking in, why does the Akka version check fail with an error message stating that an artifact is on version `2.6.0`?

ignasi35:
There's something different in the lagom build. At this point I would only compare it to the play build though. I haven't seen in detail all of the builds in this list:

> akka-http-cors, akka-http-session, akka-http, akka, play-ws, playframework, sttp...

But I suspect that the lagom build may only be compared with the playframework build. As in, both are frameworks _using_ akka artifacts. Lagom is built on top of Play. 

The Lagom uses a few more akka artifacts than the play build (cluster, management, persistence,...). 


Other than that, the only option I can think of to track the cause of why the `fail-mixed-versions` falg is ignored is using `dependencyTree` in the exact same runtime than the Jenkins build (where `-Doverride.akka.version=2.6.0` seems to kick in).

SethTisue:
@dwijnand feel free to reopen if you end up tackling this

PR #1239, opened 2020-10-02
scala-parser-combinators: track 1.2.x instead of 1.1.x (and remove scalameter for now)

SethTisue:
this was specifically prompted by a JDK 15 fix having landed on 1.2.x, but also perhaps it's just a good idea generally to be testing the newer branch?

if it doesn't work out, I can always backport the JDK 15 thing to 1.1.x (it's just an sbt-osgi version bump)

fyi @Philippus

~~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1939/~~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/1941/


Philippus (approved):

SethTisue:
two projects failed because they are still using scala.util.parsing.json, namely scalameter and lift-json, and their only downstream project is dispatch (downstream from lift-json)

I think it would be fine fine to go forward with this and eject those three projects from the build, but notify their maintainers so they can fix the problem and be reinstated if they want

oh but actually in the case of dispatch, I think we can just exclude the `dispatch-lift-json` subproject

and for lift-json, it looks like the dependency might just be in test code, so we could just not compile or run the tests

SethTisue:
@Philippus LOL, I'm just realizing that there isn't any 1.2.x release for Scalameter to upgrade to! wdyt might it be time to roll one? (I could open a ticket in the scala-parser-combinators repo)

SethTisue:
@axel22 interested in switching ScalaMeter to use a different JSON library? looks like you're only using it in one file (`HtmlReporter.scala`)

PR #1278, opened 2020-11-03
move to newer scalameta + scalafmt + scalapb + zinc

SethTisue:
this may get bogged down in version hell, but it's worth a try, now and then surprises are pleasant

~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2081/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2082/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2083/~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2090/


SethTisue:
~looks like zinc and scalameta are not going to agree about scalapb~

~possible paths forward:~
~* try harder to find a version they can both use~
~* see if I can submit an upgrade upstream~
~* split the build into spaces again~
~* eject zinc from the build and suggest they add "latest Scala nightly" to their own CI matrix~
~* keep zinc but build it using binary dependencies instead of dbuild-built dependencies?~
~* ...?~

never mind, it's something else. I'll dig into it

SethTisue:
on JDK 11,

```
[scalameta:error] Unrecognized VM option 'UseParNewGC'
[scalameta:error] Error: Could not create the Java Virtual Machine.
[scalameta:error] Error: A fatal exception has occurred. Program will exit.
[scalameta:error] --scala_out: protoc-gen-scala: Plugin failed with status code 1.
[scalameta] [error] java.lang.RuntimeException: protoc returned exit code: 1
```

though I'm not seeing the same failure locally, so this might be an outdated sbt script on our Jenkins? the usual `removeJavaOptions` trick isn't working, not sure why. I'll give `-XX:+IgnoreUnrecognizedVMOptions` a try, but I fear that the VM options are being set in some idiosyncratic way I'll have to dig into. via `JAVA_OPTS` perhaps? the sbt plugin involved is https://github.com/thesamet/sbt-protoc

~newer scalameta moves to a newer version of the plugin where apparently the way protoc is invoked changed, so one possibility is to just try the most recent tagged version of scalameta and see if the problem goes away. and if that breaks any downstream repos, probably those repos can be encouraged to upgrade~ didn't help

also doing a JDK 8 run over at https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-jdk8-integrate-community-build/3848/

SethTisue:
well this hasn't been “pleasant” exactly, but it was tractable. nearing finish line.

SethTisue:
one pleasant outcome of this PR is that the portion of the build downstream from ScalaTest 3.0 shrinks quite a bit:

```
~/cb3 % grep scalatest-3-0 dependencies.txt| wc -l
     139
~/cb3 % git switch unfreeze-scalameta
~/cb3 % grep scalatest-3-0 dependencies.txt| wc -l
      88
```

PR #1281, opened 2020-11-06
CB 2.12 fixes

dwijnand:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-jdk8-integrate-community-build/6209/console

```
Lines of Scala code recompiled during this run only:
    15305 scalafix
     2955 sbt
      362 scala-rewrites
      155 shapeless-java-records
    18777 TOTAL
SUCCEEDED: 211
FAILED: 3
BLOCKED, DID NOT RUN: 0
TOTAL: 214
```

This is good to go.

### DIFF
--- a/nightly.properties
+++ b/nightly.properties
@@ -1,2 +1,2 @@
-# November 2, 2020
-nightly=2.12.13-bin-42f2bd8
+# November 4, 2020
+nightly=2.12.13-bin-2d4985d
--- a/proj/mdoc.conf
+++ b/proj/mdoc.conf
@@ -15,6 +15,7 @@ vars.proj.mdoc: ${vars.base} {
   name: "mdoc"
   uri: "https://github.com/scalacommunitybuild/mdoc.git#23d958c5c8cbce81d59b889be7198003b39236ad"
 
+  extra.sbt-version: ${vars.sbt-1-2-version} # otherwise "NoSuchMethodError: lmcoursier.definitions.ToCoursier$.project"
   extra.exclude: [
     "lsp"  // Olaf says: "please exclude [...] it's an undocumented and untested module"
     "js", "jsdocs", "docs", "unit"  // no Scala.js plz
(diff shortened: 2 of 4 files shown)


SethTisue (commented):
do you intend to go on and fix scala-rewrites and/or sbt soon? if so, this is mergeable. if the intent is to leave them broken indefinitely, then we I'd prefer we pull them from the build entirely.

dwijnand:
Well they'll be fixed once they upgrade to 2.12.13 proper.  Or at least in theory: their dependencies need to be "2.12.13 compatible".

So I guess we'll revisit them in \<insert dafty ambitious time window\> time after 2.12.13 is out?  And merge meanwhile?

SethTisue:
am I right to suppose that scala-rewrites can and should be fixed fairly promptly after 2.12.13 comes out?

as for zinc, we're building version 1.2.1, which is reallllly old. so maybe zinc is the candidate for just dropping?

(let's keep discussing, but I'm going to hit merge now, regardless)

dwijnand:
> am I right to suppose that scala-rewrites can and should be fixed fairly promptly after 2.12.13 comes out?

The repo, yeah, but its semanticdb scalac plugin (which uses Global#reporter) might be stuck on some older version, no?

> as for zinc, we're building version 1.2.1, which is reallllly old. so maybe zinc is the candidate for just dropping?

Maybe, but as it's the only thing that actually found a bug I feel like we should give it a chance...  Let's say at least enough to figure out why it's stuck on a binary kind-projector.

SethTisue:
> The repo, yeah, but its semanticdb scalac plugin (which uses Global#reporter) might be stuck on some older version, no?

hmm, I see. I think scalameta will support 2.12.13 not long after release, so perhaps we can unfreeze that one and then if something downstream has to be removed as a result, so be it

okay, I've set a reminder to myself to look at these again after 2.12.13 is out

SethTisue:
waiting for https://repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.12.13 to land

PR #1293, opened 2020-11-13
Use latest version of Mercator 0.5.0

propensive:
@SethTisue Seeing as you drew my attention to this, I've reshuffled a few things in the repository. I hope these are correctly reflected in the changes I've made here, but chances are reasonably high I don't know what I'm doing.

### DIFF
--- a/proj/mercator.conf
+++ b/proj/mercator.conf
@@ -1,4 +1,4 @@
-// https://github.com/propensive/mercator.git#master
+// https://github.com/propensive/mercator.git#main-scala-2.13
 
 // dependency of magnolia, avro4s
 
--- a/proj/mercator.conf
+++ b/proj/mercator.conf
@@ -4,7 +4,7 @@
 
 vars.proj.mercator: ${vars.base} {
   name: "mercator"
-  uri: "https://github.com/propensive/mercator.git#c7926bc8aac564a76c64f361df1618d74861d724"
+  uri: "https://github.com/propensive/mercator.git#0f801fefd3ae7c7951f8b35a0737e8efbd8f28cd"
 
   extra.projects: ["core", "test"]
   extra.commands: ${vars.default-commands} [
--- a/proj/mercator.conf
+++ b/proj/mercator.conf
@@ -4,9 +4,9 @@
 
 vars.proj.mercator: ${vars.base} {
   name: "mercator"
-  uri: "https://github.com/propensive/mercator.git#771f7d8c7683b26864c897b8f8898dc5eeceb286"
+  uri: "https://github.com/propensive/mercator.git#c7926bc8aac564a76c64f361df1618d74861d724"
 
-  extra.projects: ["coreJVM", "tests"]
+  extra.projects: ["core", "test"]
   extra.commands: ${vars.default-commands} [
     "appendScalacOptions -Ymacro-annotations"
     "removeDependency org.scalamacros paradise"


SethTisue:
@propensive the code at the commit in question (propensive/mercator@c7926bc8aac564a76c64f361df1618d74861d724) seems to be 2.12-only code, resulting in:

```
[mercator] [error] /Users/tisue/cb3/target-0.9.17/project-builds/mercator-343f3a364e9ea88dc6d44118033d12a019bbc09b/src/core/compat.scala:26:41: type Coll takes type parameters
[mercator] [error]     }.map { xs => xs.reverseIterator.to[Coll] }
[mercator] [error] /Users/tisue/cb3/target-0.9.17/project-builds/mercator-343f3a364e9ea88dc6d44118033d12a019bbc09b/.sbt/core/../../src/core/compat.scala:26:41: type Coll takes type parameters
[mercator] [error]                                         ^
[mercator] [error]     }.map { xs => xs.reverseIterator.to[Coll] }
[mercator] [error]                                         ^
```

note the use of 2.12-style `.to[Coll]` instead of 2.13-style `.to(Coll)`

I see the same error outside of dbuild if I simply clone the mercator repo, check out that commit, run sbt, and do `++2.13.3!` followed by `compile`

propensive:
I think it should work now I'm using the correct branch (though it's a Scala 2.13-only branch).

SethTisue:
in local testing, it seems that mercator depends on probably-cli, and probably-cli depends on escritoire, and escritoire doesn't have an sbt build, I guess because you're building it with fury. I could pull an escritoire binary from Maven Central, though we generally prefer to build things from source. is adding an sbt build to escritoire something you would consider?

also: is probably-cli supposed to be a test-only dependency of mercator? the dependency isn't declared with `% Test`, but it appears to only be used in test code? one way to avoid this for now would be if you declared the dependency as test-only; then we could add mercator without compiling or running its tests; then attempting to compile and run the tests could be a separate task. but this isn't really a great route, because if we aren't running the tests, then we don't know if it works, and as a result we might get weird, hard-to-troubleshoot problems downstream instead.

and another issue: we have magnolia frozen at propensive/magnolia@5cfe85ea6fa031b1f05e3128b14b453d3af1f78f because at some point in the past, trying to use a more recent commit broke avro4s downstream. at that commit of magnolia, it expects mercator's coordinates to be `com.propensive#mercator`, but at the commit this PR adds, the dependency is instead published at `com.propensive#mercator-core`. I could override that in dbuild, which might work or might turn into a rabbit hole.

in sum: all of this would need a fair amount more work to become mergeable.

if you're interested in having Propensive-world included in the community build, I suggest that we take escritoire as the starting point, and then work our way up the dependency chain from there. but escritoire will need to have an sbt build. note that such a build can be pretty bare bones. for example, in order to build Haoyi-world, we have sbt builds for projects that Haoyi himself builds with mill, but those sbt builds can be pretty simple; see for example https://github.com/scalacommunitybuild/acyclic/blob/community-build-2.13/build.sbt (and much of this is about telling sbt about the unusual source directory layout; the minimal essentials are `name`, `organization`, and `libraryDependencies`)

propensive:
Thanks for the full explanation, @SethTisue. I've got about 20 projects, so I've been spending a lot of time over the last few weeks to try to organize their sources, migrate to Scala 2.13, publish binaries, standardise their builds (which primarily use Fury, but I intend all to have simple `sbt` builds for the purpose of publishing) and documentation, plus generally establishing my own routine for doing maintenance. That's been making steady progress, but it's a whole lot of work...

Incidentally a lot of my projects get good test coverage just by compiling them, because they're largely macros. Which is to say, it's not completely useless just to compile them, but obviously isn't as good as actually running them, of course... ;)

(I've got a spreadsheet with one column per project, and one row per task (e.g. "publish binary", though some are more fanciful, like "benchmarks") so I've added "include in Community build" in that todo list. Currently 184 out of 572 items are "done"...)

Regarding my test modules, I have always always simply _run_ my tests (which are written with the Probably test library) as applications rather than putting through sbt's test runner: if that application returns a zero exit-status, then the tests passed. I'm not sure if this is sufficient for dbuild to identify success or failure... I also have a contribution of an sbt test runner for Probably, which I'll get around to merging and publishing at some point, and would enable me to define those sbt test modules in a more traditional way.

As for Magnolia, the better solution would be to update Avro4s to work with the latest version (or if the bug is with Magnolia, fix the bug and publish 0.18). I'm making enquiries about how we can make progress here.

SethTisue:
re: running tests, I think dbuild can do that via `extra.test-tasks`. (there might be an issue where you need to make a task key, but we could cross that bridge when we come to it.) but a sbt test runner for Probably would be great.

closing this for now, happen to reopen anytime.

PR #1316, opened 2020-12-04
unfreeze elastic4s

SethTisue:
context: I want to cut down on the number of projects in the build that are still on sbt 0.13 and/or ScalaTest 3.0

and if scaladex fails downstream, perhaps I could ask someone at the Scala Center to attempt an upgrade

https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2247/


SethTisue:
this won't work — Scaladex uses elastic4s-embedded, which has been discontinued, as per https://github.com/sksamuel/elastic4s/issues/1656

breadcrumb left: 8e3cf4a5503c2c91dd27786f90266731162baced

PR #1341, opened 2021-01-14
adjustments now that Scala 2.12.13 is out

SethTisue:
Not sure if this will lead to actual changes, or whether I'll just want to update comments about where things stand.

https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-jdk8-integrate-community-build/6402/

### DIFF
--- a/proj/sbt.conf
+++ b/proj/sbt.conf
@@ -9,7 +9,7 @@
 
 vars.proj.sbt: ${vars.base} {
   name: "sbt"
-  uri: "https://github.com/scalacommunitybuild/sbt.git#de31feddf57066038e2aa54a894c66e269ac68de"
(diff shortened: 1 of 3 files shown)


SethTisue:
* sbt: we have `check-missing: false` which as usual we need for scripted tests but which (also as usual) opens us up to accidentally getting published binaries for dependencies instead of dbuild-built ones; I think that's what happened here and https://github.com/scalacommunitybuild/sbt/commit/57bb8d403287ba19686401c7da06a588ae07be1e should fix it
* scala-rewrites: needs version number bumping, now that Scalameta 4.4.6 is out with 2.12.13 support

SethTisue:
fyi @dwijnand

PR #1374, opened 2021-02-25
misc unfreezes (sbt-dotty related)

SethTisue:
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2528/~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2541/

### DIFF
--- a/proj/ip4s.conf
+++ b/proj/ip4s.conf
@@ -1,12 +1,8 @@
-// https://github.com/Comcast/ip4s.git#fbb15a58d9afc8d6fd4d3311f77ef880675b54f4  # was main
-
-// frozen (November 2020) at a commit just before Scala 3 support was added, because
-// "Missing dependency: the library org.scala-lang#scala3-library is not provided"
-// multiple repos are affected -- see remarks in fs2.conf
+// https://github.com/Comcast/ip4s.git#main
 
 vars.proj.ip4s: ${vars.base} {
   name: "ip4s"
-  uri: "https://github.com/Comcast/ip4s.git#fbb15a58d9afc8d6fd4d3311f77ef880675b54f4"
+  uri: "https://github.com/Comcast/ip4s.git#c13ad068e329d86ce1fc3829024409034cbcb85e"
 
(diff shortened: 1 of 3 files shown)


SethTisue:
They all still fail. I hypothesize that the common element here is sbt-spiewak-sonatype, which brings in an outdated sbt-dotty. I've submitted https://github.com/djspiewak/sbt-spiewak/pull/47

SethTisue:
need to wait for these to be merged, then advance the SHAs again:

* https://github.com/scodec/scodec-cats/pull/18
* https://github.com/scodec/scodec-stream/pull/135
* https://github.com/Comcast/ip4s/pull/259

SethTisue:
gah, I was so confident that the sbt-dotty 0.5.3 upgrade would fix this, but it didn't, nor does downgrading to sbt 1.4.7

there was another repo (fs2, I think it was?) where the problem went away by itself and so I assumed the sbt-dotty bump must have been what did it

I suspect that sbt-spiewak is doing some hijinks, but in itself that wouldn't explain why the fs2 repo no longer has this problem. still, I could try removing sbt-spiewak and see if the problem goes away

I'm also wondering if there might be some dependency that was published with a stray dependency on scala3-library

SethTisue:
I'm using scodec-stream to experiment locally. I've determined that if I remove sbt-spiewak from `plugin.sbt` but add back in all the plugins sbt-spiewak depends on, that fixes it. So it seems like there must be some code in sbt-spiewak itself that is causing this.

sbt-spiewak has some likely-looking hijinks in `dottyLibrarySettings`, but scodec-stream doesn't use `dottyLibrarySettings`, so I'm rather mystified.

SethTisue:
I think I'm just going to wait and see if we end up being forced to unfreeze any the 3 affected repos. fs2 was the one that really bothered me, but we aren't experiencing this there anymore

in a9e594f46564d34b37430b95d9e0bcf7880153da I updated the comments to point to this PR, so I can resume investigating if needed

PR #1377, opened 2021-03-01
2.13: advance project SHAs

SethTisue:
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2545/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2546/~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2547/

PR #1378, opened 2021-03-01
play-file-watch should work now

mkurz:
https://github.com/playframework/play-file-watch/issues/46 got fixed by https://github.com/playframework/play-file-watch/pull/119


SethTisue:
thanks for the heads up — I'm going to include this in #1377

PR #1381, opened 2021-03-06
literally add literally

SethTisue:
I thought I'd go ahead with the addition since https://github.com/Comcast/ip4s/pull/261 means we're going to need it


SethTisue:
but I hit the same scala3-library issue that also affects other sbt-spiewak based projects, so I'm going to close this for now and come back to it when something forces us to unfreeze one of those projects

fyi @mpilquist

PR #1425, opened 2021-06-07
unfreeze airframe

SethTisue:
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2853/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2856/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2857/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2860/~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2866/


SethTisue:
the timeout problem, which I noticed before but never investigated or reported upstream, [hasn't gone away](https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2857/console):

```
[airframe:error] CrockfordBase32Test:
[airframe:error] ULIDCalendarTest:
[airframe:error] ULIDTest:
[airframe:error]  - generate ULID 28.23ms
[airframe:error]  - Encode long pairs 100.08ms
[airframe:error]  - timestamp valid 124.43ms
[airframe:error]  - toString/toBytes 37.13ms
Build timed out (after 15 minutes). Marking the build as aborted.
```

@xerial any idea what this is about?

xerial:
@SethTisue airframe-ulid uses SecureRandom. It seems this random generator occasionally causes such a JVM pause. https://github.com/wvlet/airframe/pull/1721

I'll remove airframe-ulid from the community build until we find a real issue.

proj/airframe.conf
```diff
@@ -1,28 +1,15 @@
...
 vars.proj.airframe: ${vars.base} {
   name: "airframe"
-  uri: "https://github.com/wvlet/airframe.git#49158a1832af354c411a6feeb2241d0c059be885"
+  uri: "https://github.com/wvlet/airframe.git#8f186064b63913dcf104f7c321f50e2d9208a9c1"
 
   check-missing: false  // ignore missing scripted-plugin
   extra.projects: ["communityBuild"]  // no Scala.js plz
   extra.exclude: [
-    "airspecJVM"  // we use a binary, because of https://github.com/wvlet/airframe/pull/1380
+    "airspecJVM", "airspecRefJVM"  // we use binaries, because of https://github.com/wvlet/airframe/pull/1380
```
xerial: Note: Now that airspec supports Scala 3.0.0, we no longer need to use binary. We should be able to remove "airspecRefJVM" from this exclude list.

SethTisue:
@xerial https://github.com/wvlet/airframe/pull/1730 seems not to have done the trick. at https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/2860/console we see `ulidJVM` getting pulled in anyway. perhaps something still included depends on it?

xerial:
@SethTisue Ah. You are right. ulidJVM is used in other modules. Let me think about another workaround. 
By the way, is there any way to tell the project that the current environment is community build or not (e.g., an environment variable CI=xxx, etc.) ? I'm thinking about using non-secure Random for community build, which has no threading issue like SecureRandom.

SethTisue:
> is there any way to tell the project that the current environment is community build or not (e.g., an environment variable CI=xxx

we don't currently have such a variable, but in our config for airframe we can set any environment variable to any value, so if you make one up we can set it for you

SethTisue:
it's also possible for us to exclude particular tests from being run, without having to exclude the entire subproject. in the diffs for this PR you'll see that we had several such exclusions already (I'm trying to taking them out, to see if they're still needed)

xerial:
ok. Thanks. Probably, adding `ulidJVM` to `extra.exclude` is better to save your time.

SethTisue:
green on JDK 11

* JDK 8: https://scala-ci.typesafe.com/job/scala-2.13.x-jdk8-integrate-community-build/4316/ queued
* JDK 16: https://scala-ci.typesafe.com/job/scala-2.13.x-jdk16-integrate-community-build/276/ queued

SethTisue:
I forgot that we aren't even trying to make this work on JDK 8. On JDK 16 `ShellTest` still fails, so I've restored that exclusion.

SethTisue:
nice to be unfrozen and to have the configuration simplified.

xerial:
Thanks! The community build found JDK16-specific issue in advance. I'll add a CI workflow using JDK16 to airframe.

PR #1466, opened 2021-08-11
use coursier instead of sbt for Scala 3 scripts

SethTisue:
thanks to https://github.com/lampepfl/dotty/pull/13081, and using the stuff I worked out at https://github.com/coursier/coursier/issues/2058#issuecomment-896490125


SethTisue:
note that for now, the `.scala` file remains separate from the shell script that runs it. it would be nice if we only needed one file instead of two.

possible ways to get there:

* use some shell magic to have both parts in the same file and split the Scala part off at runtime for passing to the coursier invocation
* switch to Ammonite
* wait for the Standalone Scala effort to bear fruit

but meh, in the community build context this is already pretty satisfying. in other contexts the two-files thing would be more annoying, but here I don't really mind

SethTisue:
small fixup to make this work on our Jenkins: cffdd5e622f95e5debc946cfca86ad3a1517aa75

PR #1472, opened 2021-08-20
2.13: radically simplify the build

### LINKED ISSUE #1456: radically simplify the build
Dale and I were talking today and Dale was pushing a bit on whether it makes sense for the Scala 2 community build to continue in its present form.

It has come up repeatedly in the past that most of the time, we don't actually need dbuild's dependency-rewiring capability. Now that Scala 2.14 isn't happening and Scala 2 will remain binary compatible forever, it would be fine to simply build every project's HEAD using the current Scala nightly. This would catch nearly all of the same Scala regressions that the current build catches.

Where that scheme breaks down is for dependencies that use `CrossVersion.full`(/`CrossVersion.patch`), such as kind-projector, scalameta, genjavadoc, at al. You can't build cats on a nightly such as Scala 2.13.7-bin-abcd123 without first publishing kind-projector for that same nightly. You can override the dependency to just use the 2.13.6 version of kind-projector, but there are times where we actually did change something in the compiler so the previous versions of things don't work anymore, and even if we didn't, there are situations involving transitive dependencies (often involving scalameta in particular) where getting everything to use the old versions is pretty hard.

So the new idea is one I don't think we've had before: we could cut the core community build down to only build compiler plugins. Then all downstream projects would pull all of their other dependencies from Maven Central, rather than using dbuild-built binaries.

This would have the following advantages:

* It would be far less fragile and far easier to maintain.
* Included projects would be far more up-to-date, on average. We wouldn't need to freeze and/or fork so many projects.
* We could include any project we wanted, even if its dependencies aren't already in the build.

Note that if we do this, we could also consider publishing kind-projector, scalameta, genjavadoc, et al to a publicly available resolver. Then almost anyone who wants to test Scala 2 nightly builds in their own CI would be able to do so, whereas today, often they can't.

SethTisue:
* no longer giant monorepo style; instead, build has nearly flat two-level structure:
  * "core" contains only `CrossVersion.full` things (mainly compiler plugins)
  * "proj" contains everything else, each built independently and built primarily from binary dependencies (everything that isn't in "core")
* unfreeze/unfork many repos and remove workarounds that are no longer needed with the new structure
* fork a small number of additional repos, as needed
* no longer build any repos in stages (e.g. no more akka/akka-stream/akka-more, no more refined/refined-more, etc), the old way no longer has any great benefit with the new structure
* drop support for sbt 0.13
* drop support for sbt 1.3 and 1.4
* drop some less-important repos for various reasons (elastic4s, gigahorse, sksamuel-exts, play-doc, reactive-mongo, reactive-mongo-bson, scalatest-3.0, scalatestplus-*, scalatest-tests)
* update scala-ci artifactory config to current installed artifactory version (7.21.12, was some much older version)
* drop some no-longer needed resolvers from artifactory config (e.g. the bintray ones)
* add two resolvers to artifactory config that Lichess (scalachess and lila-ws) needs

fixes #1456


SethTisue:
`SUCCEEDED: 172 FAILED: 77` isn't bad

I looked at a sampling of the failures and some of them appear to involve a dependency inappropriately being looked up with full cross-versioning, rather than using the binary version as expected

I suspect that a lot of projects just need to have old, weird workarounds removed

SethTisue:
> I looked at a sampling of the failures and some of them appear to involve a dependency inappropriately being looked up with full cross-versioning, rather than using the binary version as expected

so far I have found two separate causes for this:

* the repo sets `crossScalaVersions` but does not set `scalaVersion`
  * why this would confuse dbuild, I don't know
  * workaround: set `scalaVersion` to something, anything. for now I'm doing this build-wide, at `community.conf` level, but maybe later I'll rip that out and add the workaround repo-by-repo
* sbt-dotty is in project/plugins.sbt
  * solution: remove it

perhaps these aren't the only two...

SethTisue:
squashed 76 commits into 1 :-)

SethTisue:
green run: https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-jdk11-integrate-community-build/3123/ `SUCCEEDED: 232`

PR #1493, opened 2021-10-04
readd pfps-shopping-cart

SethTisue:
it was removed because CE2 vs CE3, but with the new community build
structure it doesn't matter anymore


SethTisue:
tested locally

PR #1534, opened 2022-02-12
unfreeze finagle

SethTisue:
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/3507/


SethTisue:
still fails on JDK 17 though, like the other Twitter projects

PR #1588, opened 2022-08-10
scala-xml 2.1.0 (was 1.0.6)

SethTisue:
as fuel for the discussion at https://github.com/scala/bug/issues/12632

see also https://github.com/scala/scala/pull/10108

https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-jdk8-integrate-community-build/7565/
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-jdk8-integrate-community-build/7566/

### DIFF
--- a/community.conf
+++ b/community.conf
@@ -162,9 +162,10 @@ build += {
       {
         name:   scala-xml
         system: aether
-        // we used to build this here, but we expect it never to change again
-        // (and if it does we can always bump the version here)
-        uri:   "aether:org.scala-lang.modules#scala-xml_2.12;1.0.6"
+        // we used to build this here, but it's simpler to just use a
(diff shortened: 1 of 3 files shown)


SethTisue:
I should probably merge this regardless of what happens over in scala/scala, since any problems the upgrade uncovers are likely to be ecosystem problems we probably want to know about.

SethTisue:
there was one bad test case in scala-xml-quote (an illegal comment which is now rejected), and we're frozen at an old version of akka-persistence-jdbc which still has a test case that uses the deprecated-then-removed scala.xml.pull (the test case has since been removed on master)

so, smooth sailing!

PR #1608, opened 2022-10-26
Scalameta 4.6.0

SethTisue:
and run more tests


SethTisue:
I split this off from #1602, so it's already been tested on Jenkins. hence, merging

PR #1677, opened 2023-06-22
2.13: new Scala SHA

SethTisue:
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/4574/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/4576/~
~https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/4577/~
https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/4599/

PR #1679, opened 2023-07-03
Run all tests for Scapegoat

saeltz:
Stumbled upon this in https://github.com/scala/community-build/pull/1677. The tests should be green by now. (Otherwise I'll take another look.)


SethTisue:
thanks — I'll include this change in #1677

PR #1724, opened 2024-11-11
add fastscala

SethTisue:
tested locally

fyi @david04

two modules excluded for now; opened https://github.com/fastscala/fastscala/issues/7 about it


david04:
Hey @SethTisue , there was an issue indeed - I believe it is all good now :) thanks!

SethTisue:
@david04 yup, I was able to un-exclude the modules 4c0a8b20616c99379ed563fb2f366394bc47fe37  :+1:

PR #1737, opened 2025-11-21
Return to mainstream scala-commons

ddworak:
This version includes changes suggested in https://github.com/scala/scala-dev/issues/893


SethTisue:
thank you! (tested locally)

